PR #12, opened 2017-06-13
Add govuk branding and refactor css (Do not merge - we need a gov.uk domain)

danmitchell-:
This PR also makes use of the govuk template for the application layout file and bumps govuk gems.

### After
screen shot 2017-06-13 at 17 08 00


danmitchell-:
Closing this as we only need the font update from this branch as the other work has been done.

PR #13, opened 2017-06-19
Update suggest register form to include request or suggest

danmitchell-:
Refactor new form to include request and suggest.

Will pass this by UCD and then add links to the live site once signed off.

screen shot 2017-06-19 at 12 55 26
screen shot 2017-06-19 at 12 55 36
screen shot 2017-06-19 at 12 55 32


app/controllers/spina/new_register_controller.rb
```diff
@@ -0,0 +1,62 @@
...
+    end
+
+    def create
+      @new_register = Spina::NewRegister.new(new_register_params)
+
+      if @new_register.valid?
+        @deskproService = DeskproFeedback.new(new_register_params)
+        @response = @deskproService.send_feedback
+        if @response.status.code != 201
+          logger.debug("Suggest register ticket creation failed with status: #{@response.status} and content: #{@response.body}")
```
endamccormack: I'm pretty sure the standard log level in production is info so this log wont go into the logs most likely. Generally you would use the error level for this.

app/assets/stylesheets/application.scss
```diff
@@ -390,6 +390,10 @@ $phases: (beta, #f47738),
   margin: 0;
 }
 
+.form-hint {
+  width: 50%;
```
endamccormack: any reason you're not using the column-one-half instead for this?

app/views/spina/new_register/thanks.html.haml
```diff
@@ -0,0 +1,11 @@
+- @page_title = 'Suggest or Request a new Register'
+.container
+  .grid-row
+    .column-two-thirds
+      = link_to 'Back', spina.new_register_path, class: 'link-back'
+
+.container
```
endamccormack: I dont think there is a need for this container, could be wrong

danmitchell-: Good spot :+1:

app/assets/stylesheets/application.scss
danmitchell-: All form elements have a max width of 50% so they go inside a `column-one-whole` which means the form-hint extends beyond the inputs(because theres a lot of text) so this is my work around to ensure they are a consistent width.

app/controllers/spina/new_register_controller.rb
danmitchell-: Removing this :)

PR #14, opened 2017-07-28
Add travis config file

danmitchell-:


endamccormack:
Can we hold off merging this until the create-register-picker one is done, just to try and make sure we dont break 2 things at once if i'm wrong in my approach?

danmitchell-:
Of course

endamccormack:
Succeeded, belt ahead!

PR #16, opened 2017-09-18
Use Zendesk for support tickets

karlbaker02:
This PR removes the current Deskpro integration, switching to use Zendesk instead, in line with the 'GDS Way'. It does include support to be able to determine whether the ticket creation process was a success or not, so that we have the option to indicate this to the user after they hit submit (the current code doesn't tell the user if a problem occurred, the new code continues doing this).


app/services/zendesk_feedback.rb
```diff
@@ -0,0 +1,37 @@
+require 'zendesk_api'
+
+class ZendeskFeedback
+  @@zendesk_url = ENV['ZENDESK_URL']
+  @@zendesk_username = ENV['ZENDESK_USERNAME']
+  @@zendesk_token = ENV['ZENDESK_TOKEN']
+
+  def initialize
+    @client = ZendeskAPI::Client.new do |config|
+      config.url = @@zendesk_url
```
danmitchell-: Could we not pass the the environment variables here to avoid lines 4-6?

karlbaker02: We could indeed.

danmitchell-:
Ship it 🚢

PR #19, opened 2017-09-27
Only show register history section is content present

danmitchell-:
### Before
screen shot 2017-10-09 at 09 11 43

### After
screen shot 2017-10-09 at 09 11 19


nacnudus:
Perfect

PR #27, opened 2017-10-23
Add new records view

danmitchell-:
We need to move the history page to /history now because the register
will live at on the show view now

## End result
screen shot 2017-11-01 at 16 47 18
screen shot 2017-11-01 at 16 47 23


app/controllers/spina/registers_controller.rb
```diff
@@ -19,8 +21,21 @@ def index
       @current_phases = Spina::Register::CURRENT_PHASES
     end
 
+    def new_show
+      @register = Spina::Register.find_by_slug!(params[:id])
+
+      @register_data = @@registers_client.get_register(@register.name.parameterize, @register.register_phase)
+      @records = @register_data.get_current_records
```
karlbaker02: This should be `get_records`

app/controllers/spina/registers_controller.rb
```diff
@@ -19,8 +21,21 @@ def index
       @current_phases = Spina::Register::CURRENT_PHASES
     end
 
+    def new_show
+      @register = Spina::Register.find_by_slug!(params[:id])
+
+      @register_data = @@registers_client.get_register(@register.name.parameterize, @register.register_phase)
```
karlbaker02: This doesn't need to be an instance variable, could just be defined inside of the method

danmitchell-: I am actually using this instance variable in the view so iv'e kept it 👍

PR #28, opened 2017-10-23
Correct css class for govuk_template application layout

danmitchell-:
#### With JS enabled
screen shot 2017-10-23 at 10 57 25

#### With JS disabled
screen shot 2017-10-23 at 10 57 16


app/views/spina/registers/index.html.haml
```diff
@@ -1,6 +1,6 @@
 - @page_title = @page.seo_title
 - @page_description = @page.description
-- @body_class = 'registers-index'
+- content_for :body_classes, 'registers-index'
```
danmitchell-: The change isn't related to this PR but it needed updating.

PR #29, opened 2017-11-07
Update client library methods [DO NOT MERGE]

danmitchell-:


Gemfile
```diff
@@ -28,7 +28,7 @@ gem 'govuk_template'
 gem 'govuk_elements_rails'
 gem 'govspeak', '~> 3.4.0'
 gem 'govuk_elements_form_builder', git: 'https://github.com/ministryofjustice/govuk_elements_form_builder.git'
-gem 'openregister-ruby', git: 'https://github.com/openregister/openregister-ruby-client', tag: 'v0.2.1'
+gem 'registers-ruby-client', git: 'https://github.com/openregister/registers-ruby-client.git'
```
gidsg: I have tagged the release now, so you can use `tag: 'v0.1.0` to lock it down.

app/decorators/helpers/spina/application_helper_decorator.rb
```diff
@@ -33,21 +31,19 @@ def crest_class_name(authority)
...
 
-    def register_description(register, phase)
-      begin
-        response = HTTParty.get("https://#{register}.#{phase}.openregister.org/register.json")
-        response['register-record']['text']
-      rescue
-        "Description not found"
-      end
+    def registers_client
+      RegistersClient::RegistersClientManager.new({ cache_duration: 3600 })
```
gidsg: do we want a shorter cache time?

karlbaker02: Do you have any suggestions @gidsg ?

gidsg: dunno, how about `600`? we can always tweak this based on feedback.

PR #31, opened 2017-11-08
Add entries ui

danmitchell-:
# After

screen shot 2017-11-08 at 18 20 36

### DIFF
--- a/app/controllers/spina/registers_controller.rb
+++ b/app/controllers/spina/registers_controller.rb
@@ -41,26 +41,20 @@ module Spina
       @register = Spina::Register.find_by_slug!(params[:id])
       @register_data = @@registers_client.get_register(@register.name.parameterize, @register.register_phase)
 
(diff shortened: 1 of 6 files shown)


app/controllers/spina/registers_controller.rb
```diff
@@ -37,6 +37,36 @@ def show
...
+
+      fields = @register_data.get_field_definitions.map{|field| field[:item]['field']}
+      all_records = @register_data.get_records_with_history
+      entries_reversed = @register_data.get_entries.reverse
+
+      @entries_with_items = entries_reversed.map { |entry|
+        records_for_key = all_records[entry[:key]]
+        current_record = records_for_key.select{|record| record[:entry_number] == entry[:entry_number]}.first
+        previous_record_index = records_for_key.find_index(current_record) - 1
+        previous_record = previous_record_index >= 0 ? records_for_key[previous_record_index] : nil
```
gidsg: is there a more concise way of trying this lookup else returning `nil`?

PR #34, opened 2017-11-09
Bump rails and update readme

danmitchell-:


Gemfile
```diff
@@ -7,7 +7,7 @@ end
 
 ruby '2.4.1'
```
gidsg: should bump to ruby 2.4.2

PR #35, opened 2017-11-09
Add form to filter by record status

danmitchell-:
# After
### Default showing only current records by default `#userneed `
screen shot 2017-11-09 at 16 40 16

### Showing closed `?status=closed`
screen shot 2017-11-09 at 16 42 22

### Showing all `?status=all`
screen shot 2017-11-09 at 16 43 15

### DIFF
--- a/app/controllers/spina/registers_controller.rb
+++ b/app/controllers/spina/registers_controller.rb
@@ -30,7 +30,18 @@ module Spina
     def new_show
       @register = Spina::Register.find_by_slug!(params[:id])
       @register_data = @@registers_client.get_register(@register.name.parameterize, @register.register_phase)
-      @records = @register_data.get_records
+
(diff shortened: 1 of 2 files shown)


app/controllers/spina/registers_controller.rb
```diff
@@ -30,7 +30,18 @@ def index
     def new_show
       @register = Spina::Register.find_by_slug!(params[:id])
       @register_data = @@registers_client.get_register(@register.name.parameterize, @register.register_phase)
-      @records = @register_data.get_records
+
+      if params[:status]
+        if params[:status] == 'closed'
```
gidsg: could use a `case` statement here?

danmitchell-: 👍

PR #38, opened 2017-11-15
Fix migrations

danmitchell-:
Needed to add version number for Rails 5.1 upgrade and was missing some
translation migrations


karlbaker02:
👍

PR #39, opened 2017-11-16
Move register info out of details component

danmitchell-:
## Before
screen shot 2017-11-16 at 10 10 06
## After

screen shot 2017-11-16 at 10 09 47

### DIFF
--- a/app/views/spina/registers/show.html.haml
+++ b/app/views/spina/registers/show.html.haml
@@ -40,26 +40,14 @@
           %dd
             = link_to @register.authority, "https://www.gov.uk/government/organisations/#{@register.authority.parameterize}", target: :blank
 
+      %h2.heading-medium About this register
       %p= @register_data.get_register_definition[:item]['text']
 
-      %details{:role => "group"}
-        %summary{"aria-controls" => "details-content-0", "aria-expanded" => "false", :role => "button"}
-          %span.summary What are Registers?
-        #details-content-0.panel.panel-border-narrow{"aria-hidden" => "true"}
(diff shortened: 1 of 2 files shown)


app/views/spina/registers/show.html.haml
```diff
@@ -40,26 +40,14 @@
...
-          %span.summary How do I use a Register in my service?
-        #details-content-0.panel.panel-border-narrow{"aria-hidden" => "true"}
-          %p There are 2 ways to use Registers. You can either:
-          %ul.list-bullet
-            %li use an #{link_to "application programming interface", "https://registers-docs.cloudapps.digital/#api-documentation-for-registers"} (API) to automatically pull data from a register into your service
-            %li download the data for this Register and keep your own copy up-to-date
+      %h2.heading-medium Using this register
+      %p There are 2 ways to use Registers. You can either:
+      %ul.list-bullet
+        %li use our #{link_to "(API)", "https://registers-docs.cloudapps.digital/#api-documentation-for-registers"} to automatically pull data from a register into your service
```
gidsg: should these start with a capital letter? CC @gds-bahmady

danmitchell-: No, apparently out style for bullets is lowercase 👍

gds-bahmady: Yes, style is lowercase, but I'm not sure about the lead-in sentence "...You can either". Better would simply be "There are 2 ways to use Registers:" but I can follow up on that separately. "Up-to-date" shouldn't have hyphens, either.

PR #42, opened 2017-11-22
Increase instance memory from 256MB to 1GB

michaelabenyohai:
The instances were crashing due to out of memory exceptions. This
occurs when trying to view the new frontend pages (records and
entry timeline) for large registers such as school. I tested
increasing to 512MB but out of memory exceptions still occurred.

### DIFF
--- a/manifest.yml
+++ b/manifest.yml
@@ -1,7 +1,7 @@
 ---
 applications:
 - name: registers-frontend
-  memory: 256MB
+  memory: 1GB
   buildpack: ruby_buildpack
   instances: 2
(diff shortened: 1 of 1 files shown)


michaelabenyohai:
Yes it is definitely a bug to be. I want to do some analysis to see how big memory could get with upcoming registers etc. And also, long term, how it could require less memory. I also wonder whether PaaS has any way of monitoring instance memory.

gidsg:
@michaelabenyohai Some other teams are using this to get their metrics out: https://github.com/alphagov/paas-cf-apps-statsd we should implement this at some point.

PR #45, opened 2017-11-24
server-side search

gidsg:
Adds server side search. Needs @danmitchell- mad frontend skillz, hence **do not merge** until design is implemented.
screen shot 2017-11-24 at 13 40 24

### ** UPDATE ** Mad frontend skillz have been applied
screen shot 2017-11-27 at 11 36 53


app/views/spina/registers/show.html.haml
```diff
@@ -68,13 +68,18 @@
...
         .form-group-wrapper.js-enabled
-          = form_tag registers_path(@register.slug), method: :get, id: 'search_records' do
+
+          = form_tag register_index_path(@register.slug), method: :get, id: 'filter_records' do
             .form-group
               = label_tag 'Filter by', nil, class: 'form-label', for: 'record_status'
               = select_tag "status", options_for_select([['Existing records', 'current'], ['Closed records', 'closed'], ['Existing and closed records', 'all']], selected: params[:status]), class: "form-control", id: 'record_status'
 
-        %h3.heading-large #{@records.total_count} Records
+        %h3.heading-large #{pluralize(@records.total_count, 'Record')}
```
danmitchell-: oooo nice change 👍

config/routes.rb
```diff
@@ -3,6 +3,7 @@
 
   Spina::Engine.routes.draw do
     resources :registers, only: [:show, :index]
+    get '/registers/:id', to: 'registers#index', as: 'register_index'
```
danmitchell-: Don't think we need this route 🤔

gidsg: I added this because what we were using previously has a `.` in it so was breaking filtering and search:
```
>>  registers_path(@register.slug)
=> "/registers.country"
```
let me know if there's a better way to fix this.

gidsg: OK I figured this out now. The issue was https://stackoverflow.com/questions/5674116/path-helpers-generate-paths-with-dots-instead-of-slashes

app/controllers/spina/registers_controller.rb
```diff
@@ -34,14 +34,18 @@ def show
...
         when 'all'
-          @records = Kaminari.paginate_array(@register_data.get_records).page(params[:page]).per(100)
+          @records = paginate(@register_data.get_records)
         end
       else
-        @records = Kaminari.paginate_array(@register_data.get_current_records).page(params[:page]).per(100)
+        @records = paginate(@register_data.get_current_records)
+      end
+
+      if params[:q]
```
karlbaker02: As a user, how might I search only current or expired records with a search term?

gidsg: Right now as I understand it this is out of the scope of the requirements. You can either search all records or filter records. You can't combine them.

gidsg: OK I spoke with @elliecraven and connor and they want the filter to apply to the search, so I will pick that up.

karlbaker02: 👍

app/assets/stylesheets/modules/_search.scss
```diff
@@ -0,0 +1,40 @@
...
+    cursor: pointer;
+    height: 38px;
+    margin-left: -2px;
+    position: absolute;
+    overflow: hidden;
+    text-indent: -5000px;
+    width: 38px;
+
+    @include device-pixel-ratio() {
+      background-size: 52.5px 35px;
```
karlbaker02: Is this legal to specify a fraction of a pixel here? @danmitchell-

danmitchell-: Yes @karlbaker02

PR #48, opened 2017-11-27
Add govuk-lint to the project

fjavierm:


gidsg:
Please can you add the lint check in Travis?

fjavierm:
I will do it after I create a PR fixing all the errors. If not, we are not going to be able to build anything. Same order followed in the RMT

PR #49, opened 2017-11-27
Change the search for the update page [DO NOT MERGE - Needs FE]

fjavierm:


app/controllers/spina/registers_controller.rb
```diff
@@ -71,10 +71,32 @@ def history
 
         { current_record: current_record, previous_record: previous_record, updated_fields: changed_fields, key: entry[:key] }
       }
+
+      filter(params[:search_input]) unless params[:search_input].nil?
```
gidsg: can we call this param `q` to be consistent with the search on the records page?

fjavierm: `q` looks a little bit meaningless (I know it stands for query, but...), if we agree in something with more meaning I can change it

michaelabenyohai:
Can we add some tests for the search functionality?

fjavierm:
We should, but I though it was a card for that.

michaelabenyohai:
There is no longer a search branch as it has been merged. The card was to make tests for that code that had been merged. I want to make sure that all the code we merge has tests, so I think this PR will need tests if it is to be merged.

Can you change this PR to be merged into the `beta-assessment` branch rather than `master`? The former will deploy to the private-beta application.

PR #53, opened 2017-11-29
Change the search for the update page

fjavierm:
- History page search.
- Refactor of register page search.
- Address the cardinality _n_ problems.
- Test framework dependencies and configuration.
- Travis configuration to run tests.
- History search tests.
- Register search test.


app/controllers/spina/registers_controller.rb
```diff
@@ -73,11 +73,52 @@ def history
...
 
+    def filter(entries_mapped_with_items, filter_value)
+      entries_mapped_with_items.select { |entry|
+        entry[:key].downcase.include?(filter_value.downcase) ||
+          contained?(entry[:previous_record], filter_value) ||
+          contained?(entry[:current_record], filter_value)
+      }
+    end
+
+    def contained?(item, request_value)
```
michaelabenyohai: Please can we merge this method with the existing `search` method? They are attempting to do the same thing and it would be good to de-duplicate this logic before merging.

I realise that the existing `search` method fails for an item with cardinality `n` field, so this would also fix the cardinality `n` problem for searching on the new records page.

spec/controllers/spina/registers_controller_spec.rb
```diff
@@ -0,0 +1,157 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Spina::RegistersController, type: :controller do
+  routes { Spina::Engine.routes }
+
+  before do
+    country_data = File.open('./spec/support/country.rsf', &:read)
```
arnau: The usual way to open a file is `File.read('./spec/support/country.rsf')`.  See https://ruby-doc.org/core-2.4.0/IO.html#method-c-read

app/controllers/spina/registers_controller.rb
```diff
@@ -54,36 +56,77 @@ def history
...
+          else
+            return true if include_in_card_n?(field_value, query)
+          end
+        end
+      end
+
+      false
+    end
+
+    def include_in_card_n?(field_values, request_value)
```
gidsg: would be good to test this functionality.
Also I would rather this were named more verbosely: `included_in_cardinality_n?`

app/controllers/spina/registers_controller.rb
```diff
@@ -54,36 +56,77 @@ def history
...
-          changed_fields = fields.select { |f| current_record[:item][f] != previous_record[:item][f] }
+          changed_fields = fields.reject { |f| current_record[:item][f] == previous_record[:item][f] }
         end
 
         { current_record: current_record, previous_record: previous_record, updated_fields: changed_fields, key: entry[:key] }
-      }
+      end
 
-      @entries_with_items = paginate(entries_mapped_with_items)
+      if params[:search_input].present?
```
gidsg: I still think we should make this `q`. At least then we are consistent. Maybe if we come up with a better name then we can update it everywhere.

app/views/spina/registers/history.html.haml
```diff
@@ -19,18 +19,18 @@
...
-          %fieldset
-            .form-group
-              %label{class: "form-label", for: "search"} Search for an update
-              %input{name: "search", id: "search", class: "form-control form-control-3-4"}
+  .search-records-wrapper#search_wrapper
+    .grid-row
+      = form_tag history_path(@register.slug), method: :get, id: 'search_updates' do
+        .column-one-third
+          .records-search
+            = label_tag 'Search updates', nil, class: 'form-label', for: 'q'
```
gidsg: Frontend is expecting param `q` but backend is expecting `search_input` so it does not work.

fjavierm: The FE and BE were aligned, till someone change it. Now, everything is `q`, but still I think this name is meaningless.

gidsg: OK, but I would rather be consistent and wrong than inconsistent!

app/controllers/spina/registers_controller.rb
```diff
@@ -54,36 +56,77 @@ def history
...
 
     private
 
-    def initialize_client
-      @@registers_client ||= RegistersClient::RegistersClientManager.new({ cache_duration: 600 })
+    def filter(entries, query)
+      entries.select do |entry|
+        entry[:key].downcase.include?(query.downcase) ||
+          contained?(entry[:previous_record], query) ||
+          contained?(entry[:current_record], query)
```
michaelabenyohai: This searches all values in the before/after items for the record, not only those that have changed. The effect is that some results in the search do not actually contain the thing you have searched for in the visual "diff" (though they would if the whole of the items were displayed). For example on the updates page for the country page, search for "republic" and you will see that there is an entry for `GM` that does not actually visually display the query string.

Is this intentional? Or maybe it's part of the use case we hadn't considered yet - if so I will speak to design.

michaelabenyohai: I've also noticed that this search searched on both field-name and field-value. Is it intentional that it also searches field-name? I can see how this might make sense for the updates page (if someone is searching for entries where a particular field-name has been updated) but this does not make sense for the records page. Currently, in production (https://registers-trial.service.gov.uk/registers/country) if you search for "official" you get no results, but in this branch you get all results because all rows have a field-name called "official-name".

This may mean that we have to split the logic slightly for records search and updates search (sorry, my fault for not thinking ahead). However, I don't think searching the updates page by field-name makes sense if we are also searching the whole item (as opposed to only those that have changed), which would involve changing the logic I mentioned in the comment above.

app/views/spina/registers/history.html.haml
danmitchell-: +1 to what @gidsg said.

michaelabenyohai:
Looks good! Just wonder whether it's worth adding a couple of extra tests for the functionality in the latest commit? i.e.:

- `history` searched by field name but `show` does not
- `history` only searches on fields that have changed in the item

michaelabenyohai:
👍

PR #56, opened 2017-11-30
Add serverside search to pipeline page

danmitchell-:
We are using [Ransack](https://github.com/activerecord-hackery/ransack#ransacks-search_form_for-helper-replaces-form_for-for-creating-the-view-search-form) on this view so the syntax is slightly different from the other search views.

We should look to use ransack on the other views.


danmitchell-:
I was under the impression that we aren't touching master whilst in private beta so merging to beta-assessment. I can pull this functionality into the live site if needed.

PR #57, opened 2017-11-30
Fix records count for js pagination

danmitchell-:


michaelabenyohai:
@danmitchell- why is this to be merged into `refactor-register-pipeline-search` and not into `beta-assessment`?

danmitchell-:
Branching has gone off :-/ 

I'll fix @michaelabenyohai

PR #61, opened 2017-12-01
Hide load more when no records present

danmitchell-:
### Before
screen shot 2017-12-01 at 16 54 10

### After
screen shot 2017-12-01 at 17 14 43


karlbaker02:
Looks good for JS-enabled. When JS is disabled, the `Showing X - Y` text is incorrect, see screenshot:

gidsg:
Does it make sense to hide the table header also if there are no results?

PR #62, opened 2017-12-04
Fix build-up of register register

karlbaker02:
This PR fixes the build-up of the `@register_registers` variable in the `RegistersController`, so that the generated array contains all registers from all environments, rather than containing sub-arrays of registers of those environments. 

To the extent of my Ruby knowledge, I believe that combining registers in the way I originally thought possible (creating an array by passing in the separate args i.e. `[registers, more_registers, etc]`) and instead by calling `to_a` on each individual register group and then using the `+` operator are probably equivalent behind the scenes, but I'm happy to be corrected on this.

### DIFF
--- a/app/controllers/spina/registers_controller.rb
+++ b/app/controllers/spina/registers_controller.rb
@@ -26,7 +26,7 @@ module Spina
       beta_register_register = @@registers_client.get_register('register', 'beta').get_records
       alpha_register_register = @@registers_client.get_register('register', 'alpha').get_records
       discovery_register_register = @@registers_client.get_register('register', 'discovery').get_records
-      @register_registers = [beta_register_register, alpha_register_register, discovery_register_register]
+      @register_registers = beta_register_register.to_a + alpha_register_register.to_a + discovery_register_register.to_a
     end
 
     def show


arnau:
Well, not exactly:

```ruby
xs = [1,2,3]
ys = [4,5,6]

[xs, ys]  #=> [[1,2,3], [4,5,6]]
xs + ys  #=> [1,2,3,4,5,6]
```

To get the same flatten array using the `[]` constructor you'd have to splat the original arrays such as `[*xs, *ys]`. But unless you have a good reason for doing it this way, `+` should be the way.

karlbaker02:
Sorry, I didn't mention that you'd want call `flatten` on the array in my previous example to get a one-dimensional array.

Calling the `Array` method (https://ruby-doc.org/core-2.4.2/Kernel.html#method-i-Array) could be an alternative, but again this is calling `to_a` on the arguments you pass in. 

As for `+`, you're right that it's implemented at `Array` level (https://ruby-doc.org/core-2.4.2/Array.html#method-i-2B). I have yet to find a method to join `Enumerable`s, which is the reasoning for `to_a`-ing the individual `Enumerable`s.

PR #63, opened 2017-12-04
Fix pagination when search of filter parameters already exist

danmitchell-:
Using the `link_to_next_page` method solves the issues of ignoring other parameters when paginating


michaelabenyohai:
This works for the `show` but not `history`. Sorry - just realised the same bug exists for the updates page. Can we fix for that too?

danmitchell-:
Done @michaelabenyohai

PR #68, opened 2017-12-07
Set @registers_client as global variable

danmitchell-:
At present, there is a performance issue with us initializing `RegistersClient::RegistersClientManager` inside a helper method so I suggest we move it out on the helper file and registers controller and into the a global variable as the entire app uses it anyway.


karlbaker02:
This is fine with me for now. @gidsg do you have any opinions on this? Happy to approve otherwise.

gidsg:
👍 (once merge conflicts are resolved)

PR #73, opened 2017-12-08
move cups service name to config

gidsg:
As this value changed recently, it reminded me this should be in config rather than hardcoded in `application.rb`


danmitchell-:
@gidsg what about `manifest.yml`?

gidsg:
@danmitchell- good point, but I think as the manifest gets read before the app starts up I don't think it can read from the ruby config.

PR #78, opened 2017-12-11
Refactor search and filtering in UI

danmitchell-:
### Context
https://trello.com/c/Or0tkaoK/1592-2-register-page-updates-to-explorer-section

Update UI as per screenshots on card

screen shot 2017-12-11 at 14 58 20
screen shot 2017-12-11 at 14 58 05

### Guidance to review
Search for a record, filter records and paginate through

### DIFF
--- a/app/controllers/spina/registers_controller.rb
+++ b/app/controllers/spina/registers_controller.rb
@@ -31,9 +31,9 @@ module Spina
 
       records =
         case params[:status]
-        when 'closed'
+        when 'archived'
           @register_data.get_expired_records
-        when 'all'
+        when 'both'
           @register_data.get_records
         else
           @register_data.get_current_records
--- a/app/controllers/spina/registers_controller.rb
+++ b/app/controllers/spina/registers_controller.rb
@@ -33,7 +33,7 @@ module Spina
         case params[:status]
         when 'archived'
           @register_data.get_expired_records
-        when 'both'
+        when 'all'
           @register_data.get_records
         else
           @register_data.get_current_records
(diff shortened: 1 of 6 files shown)


app/views/spina/registers/show.js.erb
```diff
@@ -1,5 +1,5 @@
 $("#records-tbody").append("<%= j(render(partial: 'record', collection: @records)) %>");
-$("#records-count").html("Showing<strong class='text-spacing'><%= @records.count + @records.offset_value %></strong><%= 'record'.pluralize(@records.count + @records.offset_value) %>")
+$("#records-count").html('Showing <strong>1 &ndash; <%= @records.count + @records.offset_value %></strong> of <strong><%= @records.total_count %></strong> records <% if params[:q].present? %> for <strong>"<%= params[:q] %>"</strong> <% end %>')
```
gidsg: extra space here `records ` as there is another space before `for`

app/controllers/spina/registers_controller.rb
```diff
@@ -31,9 +31,9 @@ def show
 
       records =
         case params[:status]
-        when 'closed'
+        when 'archived'
```
gidsg: I think this is OK for now, as this is private beta, but in future we should refrain from changing these params even when the UI changes, as it will break bookmarked URLs.

app/controllers/spina/registers_controller.rb
```diff
@@ -31,9 +31,9 @@ def show
...
           @register_data.get_expired_records
-        when 'all'
+        when 'both'
```
gidsg: I am not sure `both` makes sense as the param name. As it would be wrong if we added more filters in future. I would rather leave this as `all`

config/locales/en.yml
```diff
@@ -30,6 +30,18 @@ en:
...
+        zero: "entries"
+        one: "entry"
+        other: "entries"
+      one_page:
+        display_entries:
+          zero: "No %{entry_name} found"
+          one: "Showing <b>1</b> %{entry_name}"
+          other: "Showing <b>all %{count}</b> %{entry_name}"
+      more_pages:
+        display_entries: "Showing <b>%{first}&nbsp;-&nbsp;%{last}</b> of <b>%{total}</b> %{entry_name}"
```
gidsg: 🌲 good to template more strings!

app/controllers/spina/registers_controller.rb
danmitchell-: +1

PR #80, opened 2017-12-11
Upgrade to registers client library v0.4.0

karlbaker02:
### Context
This PR upgrades the app to use registers client library v0.4.0, which contains some breaking changes to the interface from the previous version.

Upgrade registers client library to v0.4.0.


michaelabenyohai:
@karlbaker02 looks like we've got some failing tests

karlbaker02:
@michaelabenyohai All fixed.

PR #83, opened 2017-12-12
Remove legacy history pages and links from pipeline page

danmitchell-:
### Context
We no longer want "history" pages

Remove views, route and links

# Before
screen shot 2017-12-12 at 09 54 45

# After
screen shot 2017-12-12 at 09 54 24


michaelabenyohai:
If I click on any of the View Register links I get:

```undefined method `register_info_path'```

danmitchell-:
Ah yes @michaelabenyohai. *prompts me to add a tech debt story for an end to end test*

PR #85, opened 2017-12-13
Decouple spina from frontend

danmitchell-:
### Context
Decouple Spina from frontend.

By disabling the spina frontend routes we have the ability to create "pages" in the repo with custom routes, like any other rails app. I have included the spina frontend routes at the bottom the routes file so that we could create pages in the CMS and access them in the frontend if we wanted to.

Remove spina namespacing and routes

### Guidance to review
Compare to the current product site - https://registers.cloudapps.digital


app/controllers/new_register_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  def suggest_register
+    @new_register = Spina::NewRegister.new
+  end
+
+  def request_register
+    @new_register = Spina::NewRegister.new
+  end
+
+  def thanks
```
gidsg: what is this empty method doing?

app/views/pages/home.html.haml
```diff
@@ -1,5 +1,5 @@
-- @page_title = @page.seo_title
-- @page_description = @page.description
+- @page_title = ''
```
gidsg: do we want to populate these variables or can we delete them?

app/controllers/new_register_controller.rb
danmitchell-: It's to identify the view and route `thanks#new_registers`

PR #87, opened 2017-12-14
implement default sort

gidsg:
### Context
Previously we did not specify any default sort, so it defaulted to an order based on the order of entries in the RSF.

If `name` field exists sort by `name` else sort by primary key.

### Guidance to review
Load a register with and without `name` field and observe sort order.


gidsg:
Going to merge this now, but we should re-visit with @conordelahunty if we want to set the sorting params so that the sorting arrow shows by default.

PR #89, opened 2017-12-15
Fix gov orgs within CMS

danmitchell-:
This was broken during the upgrade to v4 of the client library.

The CMS doesn't have access to the application_controller so we need to initialize separately.


app/controllers/spina/admin/registers_controller.rb
```diff
@@ -53,6 +55,15 @@ def set_breadcrumb
         add_breadcrumb t('spina.registers.scaffold_name_plural'), spina.admin_registers_path
       end
 
+      def initialize_client
```
gidsg: Could we move this out into a shared helper method as it is identical in both usages?

PR #90, opened 2017-12-18
Show total number of records in meta info

danmitchell-:
# After
screen shot 2017-12-18 at 12 04 26

### DIFF
--- a/app/views/registers/show.html.haml
+++ b/app/views/registers/show.html.haml
@@ -36,6 +36,9 @@
           %dt Managed by:
           %dd
             = link_to @register.authority, "https://www.gov.uk/government/organisations/#{@register.authority.parameterize}", target: :blank
+          %dt No. of records:
+          %dd
+            = @register_data.get_records.count
           %dt Updated:
           %dd
             = formatted_date(Array(@register_data.get_entries).last.timestamp)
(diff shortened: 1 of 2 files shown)


app/views/registers/show.html.haml
```diff
@@ -36,6 +36,9 @@
           %dt Managed by:
           %dd
             = link_to @register.authority, "https://www.gov.uk/government/organisations/#{@register.authority.parameterize}", target: :blank
+          %dt No. of records:
```
gidsg: @bahmady is this abbreviation OK, or would you rather write out the whole word?

gidsg: @danmitchell- you should put a comma in large numbers as per: https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#numbers you can use my code from https://github.com/openregister/registers-frontend/pull/76

bahmady: @gidsg Ideally it should be "Number"

danmitchell-: Was going off the mockups from @conordelahunty but happy to update 👍

bahmady: GDS guidance on acronyms and abbreviations is a bit unclear, but something like "No." would generally be avoided because it's the sort of thing that's not easy for someone with English as a second language to parse.

danmitchell-: Updated @bahmady! Thanks for insight 👍

PR #91, opened 2017-12-18
Drop history page fields and database tables

danmitchell-:
### Context
We no longer require the history page so this PR removes the remaining code around it.

Drop database tables and code relating to the legacy history pages

### Guidance to review
pull, migrate, edit a register in the CMS to check for errors


michaelabenyohai:
I get the following when trying to edit a register

```
Showing /dev/registers-frontend/app/views/spina/admin/registers/_form.html.haml where line #51 raised:

undefined method `url' for #<Spina::Register:0x00007fe1d9ae2c58>
Did you mean?  URI
```

PR #95, opened 2017-12-18
Use local DB to render history page

karlbaker02:
### Context

This PR changes the history page to get data from the local database instance, rather than via the client library. It does this by adding `entry_number` and `previous_entry_number` to the `Entry` model. Three queries are made to the database: one query to get a set of 100 entries ordered descending on entry number, another query to get the total number of entries and the third to get the required previous entries.

Use local database to build-up the history page, rather than the client library.

### DIFF
--- a/app/controllers/registers_controller.rb
+++ b/app/controllers/registers_controller.rb
@@ -57,35 +57,66 @@ class RegistersController < ApplicationController
 
   def history
     @register = Spina::Register.find_by_slug!(params[:id])
-    @register_data = @registers_client.get_register(@register.name.parameterize, @register.register_phase)
-    fields = @register_data.get_field_definitions.map { |field| field.item.value['field'] }
-    all_records = @register_data.get_records_with_history
-    entries_reversed = @register_data.get_entries.reverse_each
 
-    entries_mapped_with_items = entries_reversed.map do |entry|
-      records_for_key = all_records.get_records_for_key(entry.key)
-      current_record = records_for_key.detect { |record| record.entry.entry_number == entry.entry_number }
-      previous_record_index = records_for_key.find_index(current_record) - 1
+    unless @register.populated
+      populate_register(@register)
+    end
 
-      if previous_record_index.negative?
+    entries = recover_entries_history(@register.id, @register.fields, params)
+    fields = get_field_definitions.map { |field| field['field'] }
+
+    @entries_with_items = entries.map do |entry_history|
+      current_record = entry_history[:current_entry]
+
+      if entry_history[:previous_entry].nil?
         changed_fields = fields
       else
-        previous_record = records_for_key[previous_record_index]
-        changed_fields = fields.reject { |f| current_record.item.value[f] == previous_record.item.value[f] }
+        previous_record = entry_history[:previous_entry]
+        changed_fields = fields.reject { |f| entry_history[:current_entry].data[f] == previous_record.data[f] }
       end
 
-      { current_record: current_record, previous_record: previous_record, updated_fields: changed_fields, key: entry.key }
+      { current_record: current_record, previous_record: previous_record, updated_fields: changed_fields, key: current_record.key }
     end
 
-    if params[:q].present?
-      filtered = filter(entries_mapped_with_items, params[:q])
-      @entries_with_items = paginate_history(filtered)
-    else
-      @entries_with_items = paginate_history(entries_mapped_with_items)
-    end
+    @entries_with_items = Kaminari::paginate_array(@entries_with_items, total_count: @page_count).page(@current_page).per(100)
   end
 
   private
+
+  def recover_entries_history(register_id, fields, params)
+    literal = params[:q]
+    page = params[:page].nil? ? 1 : params[:page].to_i
+
+    if literal.present?
+      partial = ''
+      operation_params = []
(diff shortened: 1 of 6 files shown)


app/controllers/registers_controller.rb
```diff
@@ -57,36 +57,64 @@ def show
...
-      records_for_key = all_records.get_records_for_key(entry.key)
-      current_record = records_for_key.detect { |record| record.entry.entry_number == entry.entry_number }
-      previous_record_index = records_for_key.find_index(current_record) - 1
+    unless @register.populated
+      populate_register(@register)
+    end
+
+    entries = recover_entries_history(@register.id, @register.fields, params)
+    register_data = @registers_client.get_register(@register.name.parameterize, @register.register_phase)
+    fields = register_data.get_field_definitions.map { |field| field.item.value['field'] }
```
gidsg: Could this use the `get_field_definitions` method in this controller, rather than querying the client library?

app/controllers/registers_controller.rb
```diff
@@ -57,36 +57,64 @@ def show
...
+  def recover_entries_history(register_id, fields, params)
+    literal = params[:q]
+    page = if params[:page].nil? then 1 else params[:page].to_i end
+
+    if literal.present?
+      partial = ''
+      fields.split(',').each { |field| partial += " data->> '#{field}' ilike '%#{literal}%' or" }
+      partial = partial[1, partial.length - 3] # Remove beginning space and end 'or'
+
+      query = Entry.where("spina_register_id = '#{register_id}' and entry_type = 'user' and (#{partial})").order(:entry_number).reverse_order.limit(100).offset(100 * (page - 1))
```
gidsg: rather than having these `and` conditions. I think you can use the ORM to build this query e.g. 
`where("spina_register_id: register_id, entry_type: 'user')` etc

app/controllers/registers_controller.rb
karlbaker02: It probably could if that's the way we want to go.

app/controllers/registers_controller.rb
karlbaker02: Sure.

app/controllers/registers_controller.rb
```diff
@@ -57,36 +57,67 @@ def show
...
+    previous_entries_query = Entry.where(spina_register_id: register_id, entry_number: previous_entries_numbers)
+
+    result = query.map do |entry|
+      previous_entry = previous_entries_query.select{ |previous_entry| previous_entry.entry_number == entry.previous_entry_number }.first
+      { current_entry: entry, previous_entry: previous_entry }
+    end
+
+    @page_count = count_query.length
+    @current_page = page
+    @total_pages = (@page_count / 100) + (@page_count % 100 == 0 ? 0 : 1)
```
gidsg: think you can use `(@page_count % 100).zero?` here

app/controllers/registers_controller.rb
```diff
@@ -57,36 +57,67 @@ def show
...
+    if literal.present?
+      partial = ''
+      operation_params = []
+      fields.split(',').each { |field| partial += " data->> '#{field}' ilike ? or" }
+      partial = partial[1, partial.length - 3] # Remove beginning space and end 'or'
+
+      operation_params.push(partial)
+      fields.split(',').count.times { operation_params.push("%#{literal}%") }
+
+      query = Entry.where("spina_register_id = '#{register_id}' and entry_type = 'user'").where(operation_params).order(:entry_number).reverse_order.limit(100).offset(100 * (page - 1))
```
gidsg: can change this for 
```
.where(spina_register_id: register_id, entry_type: 'user')
```

app/controllers/registers_controller.rb
```diff
@@ -57,36 +57,67 @@ def show
...
+      count_query = Entry.select(1).where("spina_register_id = '#{register_id}' and entry_type = 'user'").where(operation_params)
```
gidsg: Can change this for
```
 .where(spina_register_id: register_id, entry_type: 'user')
```

app/controllers/registers_controller.rb
karlbaker02: If we're going to change this occurrence then we should change all occurrences (including Javiers).

app/controllers/registers_controller.rb
```diff
@@ -57,36 +57,67 @@ def show
...
-      @entries_with_items = paginate_history(entries_mapped_with_items)
-    end
+    @entries_with_items = Kaminari::paginate_array(@entries_with_items, total_count: @page_count).page(@current_page).per(100)
   end
 
   private
 
+  def recover_entries_history(register_id, fields, params)
+    literal = params[:q]
+    page = if params[:page].nil? then 1 else params[:page].to_i end
```
gidsg: This simple if else could be a ternary?

app/controllers/registers_controller.rb
```diff
@@ -57,36 +57,67 @@ def show
...
+
+    if literal.present?
```
gidsg: Don't think `.present` is required in this case. `if literal` would be fine. Also as I don't *think* this is used elsewhere in the method do we need to assign a variable to it?

karlbaker02: It is pushed n times into `operation_params`, once for every field that exists in the register, so we probably should keep it inside a local variable.

PR #97, opened 2017-12-19
Extend register meta data

danmitchell-:
### Context
Give users context of the register and data

Add information to meta data section

### Guidance to review
Edit a register in the CMS, add `related_registers` and `contextual_data`, visit register on frontend and look in the meta data section  

screen shot 2017-12-19 at 12 01 37

### DIFF
--- a/app/views/registers/show.html.haml
+++ b/app/views/registers/show.html.haml
@@ -44,9 +44,20 @@
             = formatted_date(Array(@register_data.get_entries).last.timestamp)
             %strong.text-spacing -
             = link_to "See all register updates", history_path(@register.slug)
+          - if @register.related_registers.present?
+            %dt Related registers:
+            %dd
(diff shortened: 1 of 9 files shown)


app/views/registers/show.html.haml
```diff
@@ -44,9 +44,20 @@
             = formatted_date(Array(@register_data.get_entries).last.timestamp)
             %strong.text-spacing -
             = link_to "See all register updates", history_path(@register.slug)
+          - if @register.related_registers.present?
+            %dt Related registers:
+            %dd
+              != @register.related_registers
```
karlbaker02: Are you wanting to output unescaped HTML here?

danmitchell-: `!=` is the same as `.html_safe`

PR #98, opened 2017-12-20
Allow stored register data to be updated

karlbaker02:
### Context

This commit makes use of the new optional `since_entry_number` parameter in the client library when calling `get_records_with_history` and `get_metadata_records_with_history` to specify the point at which to get any new register data from. This allows the frontend to check for new records / entries on a regular basis without needing to recompute all of the stored local data about a particular register.

Use new method signature of `get_records_with_history` and `get_metadata_records_with_history` from the client library to get only updated data, allowing the local store to be kept up-to-date. 

Refactor `populate_user_data` and `populate_system_data` into a more generic `populate_data` method, which accepts an `entry_type`.

### Guidance to review

Point `registers-ruby-client` in the Gemfile to commit `4651abd327b96954fd198ff0440b7bb3c845bbd1` on GitHub, then choose a register and progressively load data into it whilst at the same time loading the `show` page of the frontend. At each point, inspect the records count for the particular register and entry type you are interested in to verify the correct data is being populated.

**This branch will fail to build as it's pointing to `master` branch of the client library** (once the client library PR is merged, this won't the case any more).

### DIFF
--- a/app/controllers/registers_controller.rb
+++ b/app/controllers/registers_controller.rb
@@ -48,9 +48,7 @@ class RegistersController < ApplicationController
   def show
     @register = Spina::Register.find_by_slug!(params[:id])
 
-    unless @register.populated
-      populate_register(@register)
-    end
+    populate_register(@register)
 
     @records = recover_records(@register.id, @register.fields, params)
   end
@@ -58,9 +56,7 @@ class RegistersController < ApplicationController
   def history
     @register = Spina::Register.find_by_slug!(params[:id])
 
-    unless @register.populated
-      populate_register(@register)
-    end
+    populate_register(@register)
 
     entries = recover_entries_history(@register.id, @register.fields, params)
     fields = get_field_definitions.map { |field| field['field'] }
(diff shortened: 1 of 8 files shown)


app/controllers/registers_controller.rb
```diff
@@ -121,56 +117,47 @@ def recover_entries_history(register_id, fields, params)
...
-          records.push(Record.new(spina_register: register, timestamp: value.entry.timestamp, hash_value: value.item.hash, entry_type: 'user', key: record[:key], data: value.item.value))
+          records.push(Record.new(spina_register: register, data: value.item.value, timestamp: value.entry.timestamp, hash_value: value.item.hash, entry_number: value.entry.entry_number, entry_type: entry_type, key: value.entry.key))
         end
 
         previous_entry_number = value.entry.entry_number
       end
 
       next unless (count % 1000).zero?
 
+      bulk_remove_existing_records(register, entry_type, records.map{ |record| record.key })
```
gidsg: Do we need to do the `bulk_remove_existing_records` and subsequent `save` as a Transaction to avoid a race condition whereby the records are deleted and the new ones not saved?

karlbaker02: Yes, this would be worth looking at.

gidsg:
Please take a look at the bulild failure on Travis.

app/controllers/registers_controller.rb
```diff
@@ -48,19 +48,15 @@ def get_field_definitions
   def show
     @register = Spina::Register.find_by_slug!(params[:id])
 
-    unless @register.populated
-      populate_register(@register)
-    end
+    populate_register(@register)
```
gidsg: since you are no longer using `register.populated` you could add a migration to remove this column.

karlbaker02: Good point.

karlbaker02: Done.

gidsg: should this column be removed from `schema.rb` also?

karlbaker02: Yep, it has been.

PR #103, opened 2017-12-28
Dockerised development environment

arnau:
### Context

PR reviews might require you to put on hold what you are doing, change environment (e.g. reset database) and after that, recover your previous state. Some other people, like me, might want to review a PR without having to install lots of dependencies at the machine level. Docker is a clean answer to solve that.

Defines a non-optimised Docker image with Ruby, node.js, Rails and the
frontend application; and a Docker Compose file with Postgres so the
only dependencies required are Docker and Docker Compose.

The Dockerfile has a hack to overwrite the database configuration so
Rails is able to connect to the dockerised Postgres.

### Guidance to review

1. Ensure you have Docker and Docker Compose installed. https://docs.docker.com/
2. `docker-compose build`
3. `docker-compose run web rails db:setup`
4. `docker-compose up -d`
5. Open `localhost:3000` on your preferred browser.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,26 @@ $ rails server
 
 You can access the CMS via http://localhost:3000/admin
 
+## Running with Docker
+
+Instead of the above, you can use [Docker][https://docs.docker.com] as
+follows:
+
+```bash
+$ docker-compose build
+$ docker-compose run --rm web rails db:setup
+$ docker-compose up -d
+```
+
+And access it normally via http://localhost:3000
+
+To stop it, run `docker-compose stop`.
+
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -24,6 +24,26 @@ $ rails server
 
 You can access the CMS via http://localhost:3000/admin
 
+## Running with Docker
+
+Instead of the above, you can use [Docker][https://docs.docker.com] as
```
gidsg: The URL should have normal parenthesis. 
`[Docker](https://docs.docker.com)`

PR #105, opened 2018-01-02
Use ActiveJob to populate Records & Entries in  DB

gidsg:
### Context
Previously we were only populating the DB when certain controller methods were hit. This had a number of downsides including slow loads on first request and inability to apply updates after first load.

Move all database population to an ActiveJob. Note this PR only triggers the job at upon startup. The intention is that a subsequent PR will change this to a scheduled job and introduce a queuing backend. 

### Guidance to review
All registers which are defined in the CMS should be loaded upon startup. Any registers that fail to load e.g. Charity should not prevent other registers loading.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -23,10 +23,10 @@ gem 'cocoon'
 gem 'haml-rails'
 
 # GDS Frontend Toolkit, templates and elements
+gem 'govuk_elements_form_builder', git: 'https://github.com/ministryofjustice/govuk_elements_form_builder.git'
+gem 'govuk_elements_rails'
 gem 'govuk_frontend_toolkit'
 gem 'govuk_template'
-gem 'govuk_elements_rails'
-gem 'govuk_elements_form_builder', git: 'https://github.com/ministryofjustice/govuk_elements_form_builder.git'
 gem 'registers-ruby-client', git: 'https://github.com/openregister/registers-ruby-client.git', tag: 'v0.6.0'
 
 # Spina CMS
--- a/app/jobs/populate_register_data_in_db_job.rb
+++ b/app/jobs/populate_register_data_in_db_job.rb
@@ -49,7 +49,8 @@ class PopulateRegisterDataInDbJob < ApplicationJob
         previous_entry_number = value.entry.entry_number
       end
 
-      next unless (count % 1000).zero?
+      next unless (count / 1000).positive?
+      
       if latest_entry_number.positive?
         Record.transaction do
           bulk_remove_existing_records(register, entry_type, records.map(&:key))
--- a/app/jobs/populate_register_data_in_db_job.rb
+++ b/app/jobs/populate_register_data_in_db_job.rb
@@ -1,0 +1,82 @@
+class PopulateRegisterDataInDbJob < ApplicationJob
+  queue_as :default
+
+  def initialize
+    @registers_client ||= RegistersClient::RegisterClientManager.new(cache_duration: 60)
+  end
+
+  def bulk_remove_existing_records(register, entry_type, record_keys)
+    unless record_keys.empty?
+      Record.where(spina_register_id: register.id, entry_type: entry_type, key: record_keys).delete_all
+    end
+  end
+
+  def bulk_save(entries, records)
+    Entry.import!(entries)
+    Record.import!(records)
+  end
+
+  def populate_register(register)
+    register_data = @registers_client.get_register(register.name.parameterize, register.register_phase.downcase)
+    register_data.refresh_data
+    populate_data(register, register_data, 'user')
+    populate_data(register, register_data, 'system')
+
+    register.fields = register_data.get_field_definitions.map { |field| field.item.value['field'] }.join(',')
+    register.save
+  end
+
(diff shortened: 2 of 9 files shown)


app/jobs/populate_register_data_in_db_job.rb
```diff
@@ -0,0 +1,117 @@
...
+      new_entry = Entry.new(spina_register: register, data: record.item.value, timestamp: record.entry.timestamp, hash_value: record.entry.hash, entry_type: 'system', key: record.entry.key)
+      entries.push(new_entry)
+      records.push(Record.new(spina_register: register, data: record.item.value, timestamp: record.entry.timestamp, hash_value: record.entry.hash, entry_type: 'system', key: record.entry.key))
+    end
+
+    bulk_save(entries, records)
+  end
+
+  def perform(*)
+    Spina::Register.all.each do |register|
```
danmitchell-: Would `.find_each` be more performant than `.all.each`

Gemfile
```diff
@@ -27,7 +27,7 @@ gem 'govuk_frontend_toolkit'
 gem 'govuk_template'
 gem 'govuk_elements_rails'
 gem 'govuk_elements_form_builder', git: 'https://github.com/ministryofjustice/govuk_elements_form_builder.git'
-gem 'registers-ruby-client', git: 'https://github.com/openregister/registers-ruby-client.git', tag: 'v0.4.0'
+gem 'registers-ruby-client', git:'https://github.com/openregister/registers-ruby-client.git',  ref: 'd9055bf52af176275c93e924ce22dcef1212708b'
```
danmitchell-: Can you just reference `branch: 'master'` and run `bundle update registers-ruby-client`

app/jobs/populate_register_data_in_db_job.rb
```diff
@@ -0,0 +1,117 @@
+class PopulateRegisterDataInDbJob < ApplicationJob
+  queue_as :default
+
+  def initialize
+    @registers_client ||= RegistersClient::RegisterClientManager.new(cache_duration: 60)
+  end
+
+  def bulk_remove_existing_records(register, entry_type, record_keys)
+    unless record_keys.empty?
+      Record.where(spina_register_id: register.id, entry_type: entry_type, key: record_keys).delete_all
```
danmitchell-: Should we use `destroy_all ` instead? https://apidock.com/rails/ActiveRecord/Relation/destroy_all

app/jobs/populate_register_data_in_db_job.rb
```diff
@@ -0,0 +1,117 @@
...
+
+    register.fields = register_data.get_field_definitions.map { |field| field.item.value['field'] }.join(',')
+    register.save
+  end
+
+  def populate_data(register, register_data, entry_type)
+    entries = []
+    records = []
+    count = 0
+    latest_entry = Entry.where(spina_register_id: register.id, entry_type: entry_type).order(:entry_number).reverse_order.limit(1).first
```
danmitchell-: Replace `limit(1).first` with `first(1)`  https://apidock.com/rails/ActiveRecord/FinderMethods/first ?

config/initializers/job_runner.rb
```diff
@@ -0,0 +1,3 @@
+unless Rails.env.test? || !ActiveRecord::Base.connection.table_exists?('spina_registers') || File.basename($0) == 'rake'
+  PopulateRegisterDataInDbJob.perform_now
```
danmitchell-: Could we move this into `production.rb` instead of having the conditional clause?

app/jobs/populate_register_data_in_db_job.rb
karlbaker02: `destroy_all` calls `destroy` on associations right (i.e. anything that's linked to the thing you're deleting by a foreign key). There are no relationships in this DB as far as I'm aware (entries and records store the item inside of themselves).

config/initializers/job_runner.rb
gidsg: Actually we want to run this at the moment on `development` also. It was just breaking tests and db migration if run in those circumstances I will see if I can refactor this, it will be replaced by a proper scheduled task in due course.

app/jobs/populate_register_data_in_db_job.rb
gidsg: Actually there is no need for the argument as it defaults to 1. Good suggestion!

app/jobs/populate_register_data_in_db_job.rb
```diff
@@ -0,0 +1,90 @@
...
+  def populate_data(register, register_data, entry_type)
+    entries = []
+    records = []
+    count = 0
+
+    latest_entry = Entry.where(spina_register_id: register.id, entry_type: entry_type).order(:entry_number).reverse_order.first
+    latest_entry_number = latest_entry.nil? ? 0 : latest_entry.entry_number
+
+    (entry_type == 'user' ? register_data.get_records_with_history(latest_entry_number) : register_data.get_metadata_records_with_history(latest_entry_number)).each do |record|
+      previous_entry_number = nil
```
arnau: Why do you need this?

PR #107, opened 2018-01-02
Add google-site-verification meta tag

danmitchell-:
Required for Google Search Console

### DIFF
--- a/app/views/layouts/default/application.html.haml
+++ b/app/views/layouts/default/application.html.haml
@@ -24,6 +24,7 @@
 - content_for :head do
   = stylesheet_link_tag 'application'
   = csrf_meta_tags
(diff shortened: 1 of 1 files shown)


app/views/layouts/default/application.html.haml
```diff
@@ -24,6 +24,7 @@
 - content_for :head do
   = stylesheet_link_tag 'application'
   = csrf_meta_tags
+  %meta{name:'google-site-verification', content: 'IJ5HOXpZrISM6la-YO_iX0rBUC5YFftpexygcKLsNs4'}
```
gidsg: should this value be coming from config?

danmitchell-: Not IMO because it will never change, its a one time thing to verify site ownership on Google

PR #116, opened 2018-01-15
Add queue and scheduler services to populate DB

gidsg:
### Context
Previously we were populating the DB only when certain views were requested. This meant there could be a long delay for the first viewer of a register. Also updates to a register were not made after the initial population

* Introduces a separate queue service using `delayed_job`
* Introduces a separate scheduled job service using `clock`
* This is broadly based on the approach described in this [Pivotal Cloud Foundry blog post](https://content.pivotal.io/blog/scheduling-tasks-on-cloud-foundry) 
* Locally a rake task is used to populate the DB when required and this is documented in the README.

### Guidance to review
* Locally: Running rake task described in README should populate DB.
* CF: Database should be regularly populated by the queue and scheduler services.


manifest.yml
```diff
@@ -13,3 +13,23 @@ applications:
     - registers-frontend
     # environment variables (persisted for blue/green deploys)
     - registers-product-site-environment-variables
+- name: registers-frontend-queue
+  memory: 1GB
+  instances: 1
+  command: bundle exec 'rake environment jobs:work'
```
danmitchell-: Does this work okay having another `command` on line 29?

app/views/registers/show.html.haml
```diff
@@ -54,10 +54,7 @@
               != @register.contextual_data
 
       %h2.heading-medium About this register
-      - if @register.register_phase == 'Backlog'
-        = @register.description
-      - else
-        %p= get_register_definition(@register.id, "register:#{params[:id]}")['text']
+      %p=get_register_definition['text']
```
danmitchell-: I think this change is wrong

app/models/spina/register.rb
```diff
@@ -9,6 +9,12 @@ class Register < ActiveRecord::Base
     validates_uniqueness_of :name
     validates :slug, uniqueness: true
 
+    has_many :steps, -> { order('position ASC') }
```
danmitchell-: We don't need these associations

manifest.yml
fjavierm: They are separate artefacts. Every `- name:` is a different app, then there is one commant for each artefact.

app/views/registers/show.html.haml
gidsg: This is required because the corresponding method has changed to read from the database record table (derived from the RSF) in: https://github.com/openregister/registers-frontend/blob/delayed-scheduler-backend/app/controllers/registers_controller.rb#L30

manifest.yml
gidsg: This PR changes the manifest to deploy multiple separate apps in CF as per: https://docs.cloudfoundry.org/devguide/deploy-apps/manifest.html#multi-apps
They are all sharing the same artefact and services but with different `command` entry points. You can see this setup deployed in `sandbox`.

lib/clock.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+require 'clockwork'
+require File.expand_path('../../config/boot',        __FILE__)
+require File.expand_path('../../config/environment', __FILE__)
+
+include Clockwork
+
+handler do |job|
```
karlbaker02: Could be an idea to include the second argument `time` and print this out to the Rails logger too, that way if we have any problems there might be more of a chance of figuring out when it started. (see the first example in the section https://github.com/Rykian/clockwork#quickstart).

manifest.yml
```diff
@@ -1,15 +1,26 @@
...
+  health-check-http-endpoint: /health_check/standard
+- name: registers-frontend-queue
+  <<: *defaults
+  instances: 1
+  command: bundle exec 'rake environment jobs:work'
+  health-check-type: process
+- name: registers-frontend-scheduler
+  <<: *defaults
+  instances: 1
+  command: bundle exec clockwork 'lib/clock.rb'
```
karlbaker02: Given that this is the instance downloading all the RSF, parsing it out and then updating the DB, it might need more than 1GB RAM, but this is something we can update later on if it's not appropriate at the moment.

app/services/populate_database.rb
```diff
@@ -0,0 +1,95 @@
...
+  end
+
+  def bulk_save(entries, records)
+    Entry.import!(entries)
+    Record.import!(records)
+  end
+
+  def populate_register
+    Delayed::Worker.logger.info("Updating #{@register.name} in database")
+    register_data = @registers_client.get_register(@register.name.parameterize, @register.register_phase.downcase)
```
karlbaker02: You're going to have a problem here, in that getting each register is going to download the entire contents of the register every time the job is invoked. This may be unavoidable, but it's because a new `RegisterClientManager` is being constructed each time the job is run, so every time it believes it hasn't seen any data before and gets the whole register again. If we could store the `RegisterClientManager` as we were before in a class variable and have the same instance 
 accessible each time the job is run, we can avoid this.

app/services/populate_database.rb
```diff
@@ -0,0 +1,95 @@
+# frozen_string_literal: true
+
+class PopulateDatabase
+  def initialize(register)
+    @registers_client ||= RegistersClient::RegisterClientManager.new(cache_duration: 60)
```
karlbaker02: This will always be a new `RegisterClientManager` because `PopulateRegisterDataInDbJob` creates a new `PopulateDatabase` instance when the `perform` method is called.

app/services/populate_database.rb
```diff
@@ -0,0 +1,95 @@
...
+            bulk_save(entries, records)
+          end
+        else
+          bulk_save(entries, records)
+        end
+      end
+    rescue StandardError => e
+      Delayed::Worker.logger.error("Problem found populating #{register.name}: #{e.message}")
+
+      if latest_entry.nil?
```
karlbaker02: Would we want to either try again or roll back the DB to the previous state? If we delete all then we're in a worse position than when we started.

bin/deploy
```diff
@@ -1,15 +1,18 @@
...
 
-echo "Deploying ${APP_NAME} to ${CF_ORG}/${CF_SPACE}..."
+for APP_NAME in $APP_ARRAY
+do
+  echo "Deploying ${APP_NAME} to ${CF_ORG}/${CF_SPACE}..."
 
-# $CF env variables are set by Travis and configured in ../.travis.yml
-cf login -a $CF_API -u $CF_USERNAME -p $CF_PASSWORD -o $CF_ORG -s $CF_SPACE
+  # $CF env variables are set by Travis and configured in ../.travis.yml
+  cf login -a $CF_API -u $CF_USERNAME -o $CF_ORG -s $CF_SPACE
```
karlbaker02: Do we still not need the password to login?

config/initializers/delayed_job_config.rb
```diff
@@ -0,0 +1,14 @@
+Delayed::Worker.destroy_failed_jobs = false
+Delayed::Worker.sleep_delay = 60
+Delayed::Worker.max_attempts = 3
+Delayed::Worker.max_run_time = 30.minutes
+Delayed::Worker.read_ahead = 10
+Delayed::Worker.default_queue_name = 'default'
+Delayed::Worker.delay_jobs = !Rails.env.test?
+Delayed::Worker.raise_signal_exceptions = :term
+
+Delayed::Worker.logger ||= if defined?(Rails)
```
karlbaker02: What happens if neither?

config/initializers/registers_client.rb
```diff
@@ -1 +0,0 @@
-@@registers_client ||= RegistersClient::RegisterClientManager.new(cache_duration: 600)
```
karlbaker02: Related to my other comment in the PopulateDatabase, we may need to add this back in so that we don't re-download all register data each time the job is run.

Gemfile.lock
```diff
@@ -163,11 +171,7 @@ GEM
     formatador (0.2.5)
     globalid (0.4.1)
       activesupport (>= 4.2.0)
-    globalize (5.1.0.beta2)
```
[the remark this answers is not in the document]
gidsg: was not intentionally removed. @danmitchell- what do you think about this one?

lib/clock.rb
gidsg: I could, but we will get this I believe from the timestamp from the log line in any case.

manifest.yml
gidsg: Actually I believe the `scheduler` just adds messages to the queue see https://github.com/Rykian/clockwork#use-with-queueing. It should need little memory and the `queue` instance should need more as it actually performs the job. It is easy to tweak this in future on a per-app basis.

app/services/populate_database.rb
gidsg: we only do this when `latest_entry.nil?` so there is no existing data for the register. When the subsequent exception is thrown the queue backend will enqueue a retry attempt automatically.

karlbaker02: Got it.

manifest.yml
karlbaker02: Sorry, this is what I meant, I put the comment on the wrong line. Sounds good.

lib/tasks/populate_register_data_in_db.rake
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
```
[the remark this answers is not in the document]
gidsg: No, this is handled by the migration script included in the PR.

config/initializers/delayed_job_config.rb
gidsg: this is fixed in https://github.com/openregister/registers-frontend/pull/116/commits/dc5fa4f17f1d9ba570e161ae02f5de6d6cce49db

Gemfile.lock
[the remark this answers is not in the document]
danmitchell-: Spina now uses `Mobility` instead of `Globalize`

app/views/registers/show.html.haml
danmitchell-: 👍

PR #118, opened 2018-01-17
fix pagination offset counts on show page

gidsg:
### Context
Previously the offset counts were incorrect as this was not implemented when we switched to an ActiveRecord model. 

Restore offset counts. Note as we are not using Kaminari to produce the total count (for performance reasons) we have do some manual calculations to work the values out.

__Non-JS__
screen shot 2018-01-17 at 09 58 44

__JS__
screen shot 2018-01-17 at 09 59 31

### Guidance to review
Offsets should work as-per `beta-assessment` branch.

### DIFF
--- a/app/controllers/registers_controller.rb
+++ b/app/controllers/registers_controller.rb
@@ -64,7 +64,7 @@ class RegistersController < ApplicationController
     @entries_with_items = Kaminari::paginate_array(@entries_with_items, total_count: @page_count).page(@current_page).per(100)
   end
 
-  private
+private
 
 
   def recover_entries_history(register_id, fields, params)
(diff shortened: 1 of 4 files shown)


app/controllers/registers_controller.rb
```diff
@@ -87,11 +87,11 @@ def recover_entries_history(register_id, fields, params)
       count_query = Entry.select(1).where(spina_register_id: register_id, entry_type: 'user')
     end
 
-    previous_entries_numbers = query.reject{ |entry| entry.previous_entry_number.nil? }.map{ |entry| entry.previous_entry_number }
+    previous_entries_numbers = query.reject { |entry| entry.previous_entry_number.nil? }.map(&:previous_entry_number)
```
gidsg: these changes are unrelated, they were automatically applied by rubocop

michaelabenyohai:
Should this be merged into the postgres branch rather than the scheduler branch?

gidsg:
Yes, @michaelabenyohai I will close and open a new PR against `postgres_based_performance`

PR #128, opened 2018-01-19
Add tracking to header title, links and breadcrumb

danmitchell-:
### Context
As part of the GaaP programme we want to be able to track more events 

Add the generic script from https://github.com/alphagov/gaap-analytics

### Guidance to review
Check data attributes match ORJ

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -13,3 +13,5 @@
 //= require jquery
 //= require jquery_ujs
 //= require_tree .
+
+window.GAAP.analytics.init();
--- a/app/views/layouts/application.html.haml
+++ b/app/views/layouts/application.html.haml
@@ -5,7 +5,7 @@
 - content_for :after_header do
   %header{class: "header #{current_page?(root_path) ? 'header--without-border' : ''}"}
     .header__container
-      .header__brand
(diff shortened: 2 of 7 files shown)


app/views/layouts/application.html.haml
```diff
@@ -30,9 +30,9 @@
             = link_to "Documentation", "https://registers-docs.cloudapps.digital/", target: :blank, rel: "noopener noreferrer"
 
 - content_for :head do
-  = stylesheet_link_tag 'application'
+  = stylesheet_link_tag 'application' if Rails.env.production?
```
karlbaker02: Should this be on the `layouts/analytics` partial still?

app/assets/javascripts/application.js
```diff
@@ -13,3 +13,5 @@
 //= require jquery
 //= require jquery_ujs
 //= require_tree .
+
+window.GAAP.analytics.init();
```
karlbaker02: Do we need to require the `gaap-analytics.js` file in here too?

danmitchell-: `//= require_tree` brings in everything from the application directory

danmitchell-: You can remove `//= require_tree .` and specify the files individually when you want to control the load order.

PR #130, opened 2018-01-19
Switch to ActiveRecord based model

gidsg:
### Context
The previous in-memory model was not performant for large registers and required a large amount of memory on the instance

Switch to a Postgres/ActiveRecord based model. Use a separate service to populate the database using the client library

### Guidance to review
This is a refactor. It should be functionally equivalent but faster and use less memory.


app/controllers/registers_controller.rb
```diff
@@ -13,71 +17,140 @@ def index
...
+      fields.split(',').count.times { operation_params.push("%#{literal}%") }
+
+      query = query.where(operation_params)
+      count_query = query.where(operation_params)
+    end
+
+    if sort_by.present? && sort_direction.present?
+      query = query.order("data->> '#{sort_by}' #{sort_direction.upcase} nulls last")
+    end
+    @page_count = count_query.count
```
arnau: 👌

app/controllers/registers_controller.rb
```diff
@@ -108,42 +108,38 @@ def recover_records(register_id, fields, params, page_size = 100)
...
-    count_query = Record.select(1).where(spina_register_id: register_id, entry_type: 'user')
 
     case status
     when 'archived'
       query = query.where("data->> 'end-date' is not null")
-      count_query = query.where("data->> 'end-date' is not null")
     when 'current'
       query = query.where("data->> 'end-date' is null")
-      count_query = query.where("data->> 'end-date' is null")
     end
```
arnau: The conventional way to make this is putting the assignment outside the case:

```ruby
query = case status
when 'archived'
  query.where("data->> 'end-date' is not null")
when 'current'
  query.where("data->> 'end-date' is null")
end
```

Also, what happens if `status` is neither "archived" nor "current"?

arnau: Ah makes sense 👍

PR #132, opened 2018-01-25
Add govuk lint

danmitchell-:
### Context
We want to lint our code, set by `govuk-lint`

Add govuk-lint to our CI pipeline. 

### Guidance to review
Run the `bundle exec govuk-lint-ruby app lib spec` to check for warnings, run `rake` to check specs all pass, then have a play with the UI


arnau:
This PR and previous conversations around Ruby 2.5 not being available makes me think that next time we need to be a bit more conservative so we can move to the next version when the rest of the stack is ready for it.

danmitchell-:
Thats expected behaviour of the gem. You can't run `bundle exec govuk-lint-sass` without the directory path

PR #136, opened 2018-01-29
check database for previous entry number

gidsg:
### Context
Previously when doing an incremental update we were only looking in the in-memory store for the `previous_entry_number`. 

If there are existing entries, look in the database and the in-memory store for `previous_entry_number`

### Guidance to review
running `rake registers_frontend:populate_db:fetch_now`  on an existing DB should populate previous entry number, for entries have a previous version in the database.

### DIFF
--- a/app/services/postgres_data_store.rb
+++ b/app/services/postgres_data_store.rb
@@ -9,7 +9,7 @@ class PostgresDataStore
     @entries = { user: [], system: [] }
     @records = { user: {}, system: {} }
     @register = register
-    @has_existing_entries_in_db = Entry.where(spina_register_id: @register.id).exists?
+    @has_existing_entries_in_db = Entry.where(register_id: @register.id).exists?
   end
 
   def add_item(item)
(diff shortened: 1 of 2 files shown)


app/services/postgres_data_store.rb
```diff
@@ -18,7 +18,16 @@ def add_item(item)
   def append_entry(entry)
     entry_type = entry.entry_type.to_sym
     item = @items[entry.item_hash]
-    previous_entry_number = @records[entry_type].has_key?(entry.key) ? @records[entry_type][entry.key].last[:entry_number] : nil
+    previous_entry_number_from_memory = @records[entry_type].key?(entry.key) ? @records[entry_type][entry.key].last[:entry_number] : nil
```
karlbaker02: I'd be tempted to make this even simpler and say something along the lines of:

```
previous_entry_number = @records[entry_type].key?(entry.key) ? @records[entry_type][entry.key].last : Record.where(register_id: @register.id, entry_type: entry.entry_type.to_s, key: entry.key.to_s).first
```

Then you could say:
```
unless previous_entry_number.nil?
  previous_entry_number = previous_entry_number[:entry_number]
end
```

or something to that avail...

PR #141, opened 2018-01-30
Update translation migration from spina

danmitchell-:
When running `rake db:setup` it was failing because of missing migrations from `spina`. This PR fixes that. 

To test -
`rake db:drop`
`rake db:setup`


danmitchell-:
Thanks @karlbaker02, did you check this @gidsg ?

gidsg:
I have not tested this @danmitchell- but if @karlbaker02 then merge it!

PR #142, opened 2018-01-30
Add unique constraints

gidsg:
### Context
Previously we had no constraints within the database to prevent the creation of duplicate entries

add unique index on `hash_value entry_type timestamp spina_register_id` 

### Guidance to review
`rake db:migrate` should add index to DB.


arnau:
@michaelabenyohai @karlbaker02 @danmitchell- this is something everyone should validate it's not going to backfire

db/migrate/20180130160640_add_unique_constraint_to_entries.rb
```diff
@@ -0,0 +1,5 @@
+class AddUniqueConstraintToEntries < ActiveRecord::Migration[5.1]
+  def change
+    add_index(:entries, %i[hash_value entry_type timestamp spina_register_id], unique: true, name: 'unique_entry_index')
```
karlbaker02: Technically you could have multiple entries created at the same time for the same register and which point at the same object. The index should instead be on `[hash_value entry_type entry_number spina_register_id]`, as `entry_number` is guaranteed to be unique for each `entry_type`.

db/migrate/20180130160640_add_unique_constraint_to_entries.rb
```diff
@@ -1,5 +1,5 @@
 class AddUniqueConstraintToEntries < ActiveRecord::Migration[5.1]
   def change
-    add_index(:entries, %i[hash_value entry_type timestamp spina_register_id], unique: true, name: 'unique_entry_index')
+    add_index(:entries, %i[hash_value entry_type entry_number spina_register_id], unique: true, name: 'unique_entry_index')
```
gidsg: I have a slight concern that this still leaves the possibility open to create a duplicate entry but with a different entry number.

db/migrate/20180130160640_add_unique_constraint_to_entries.rb
arnau: A bit late but, I'm missing `key` here. If `key` is part of the index, isn't unique enough to make `timestamp` relevant?

db/migrate/20180130160640_add_unique_constraint_to_entries.rb
karlbaker02: I'm not sure I understand what you mean here by a duplicate entry - could you elaborate please?

gidsg: where the other fields are the same but the entry number has been mis-calculated. As we have seen examples of this.

db/migrate/20180130160640_add_unique_constraint_to_entries.rb
```diff
@@ -0,0 +1,5 @@
+class AddUniqueConstraintToEntries < ActiveRecord::Migration[5.1]
+  def change
+    add_index(:entries, %i[hash_value entry_type entry_number spina_register_id key], unique: true, name: 'unique_entry_index')
```
danmitchell-: Should be `register_id` now

PR #144, opened 2018-02-01
Remove spina admin layout overrides and refactor gov orgs helper

danmitchell-:
### Context
Cleaning up the amount of Spina code in our repo

Remove spina layout overrides from codebase

### Guidance to review
Login into the CMS and edit a register


app/controllers/spina/admin/registers_controller.rb
```diff
@@ -15,6 +16,10 @@ def new
         @register = Spina::Register.new
       end
 
+      def edit
+        add_breadcrumb @register.name
```
gidsg: is it convention to use brackets for the args?

danmitchell-: Good spot!

app/views/spina/admin/registers/_form.html.haml
```diff
@@ -64,7 +64,9 @@
   - unless @register.new_record?
     .pull-right= link_to t('spina.permanently_delete'), spina.admin_register_path(@register), method: :delete, data: {confirm: t('spina.pages.delete_confirmation', subject: @register.name) }, class: 'button button-link button-danger'
 
-= content_for :head do
+= content_for :plugin_stylesheets do
+  = javascript_include_tag '//unpkg.com/accessible-typeahead@0.4.0', data: {turbolinks_track: true}
```
gidsg: would be nice if we could host this ourselves using our own build pipeline, but not a blocker.

PR #146, opened 2018-02-01
Update docs link to test site for UR

danmitchell-:
### Context
We are doing UR on the tech docs and to complete the user journey from the product page we want to use the test environment URL.

Update docs link to point to test site

### Guidance to review
Check link in header

### DIFF
--- a/app/views/layouts/application.html.haml
+++ b/app/views/layouts/application.html.haml
@@ -27,7 +27,7 @@
           %li{class: "#{'active' if current_page?(support_path)}"}
             = link_to "Support", support_path
           %li
-            = link_to "Documentation", "https://registers-docs.cloudapps.digital/", target: :blank, rel: "noopener noreferrer"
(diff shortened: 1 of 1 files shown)


app/views/layouts/application.html.haml
```diff
@@ -27,7 +27,7 @@
           %li{class: "#{'active' if current_page?(support_path)}"}
             = link_to "Support", support_path
           %li
-            = link_to "Documentation", "https://registers-docs.cloudapps.digital/", target: :blank, rel: "noopener noreferrer"
+            = link_to "Documentation", "https://registers-docs-testing.cloudapps.digital/", target: :blank, rel: "noopener noreferrer"
```
gidsg: I guess this is OK, but ideally we'd push this to the sandbox instance of registers-frontend so we don't have to modify production just for user testing.

danmitchell-: Agreed. It's a temp solution whilst registers-trial is locked behind http auth.

PR #149, opened 2018-02-01
only show registers in ready to use for which we have data

gidsg:
### Context
Previously if register population failed we would still show a link to the register on the homepage

Only show a link if we have Records in the DB for the register

### Guidance to review
Links should only show for registers where records exist in DB.

### DIFF
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -22,6 +22,6 @@ class PagesController < ApplicationController
 private
 
   def get_ready_to_use_registers
-    @beta_registers = Register.where(register_phase: 'Beta').select { |r| Record.where(register_id: r.id).exists? }
+    @beta_registers = Register.where(register_phase: 'Beta').select(&:records?)
   end
 end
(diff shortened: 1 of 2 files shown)


app/models/spina/register.rb
```diff
@@ -12,6 +12,10 @@ class Register < ActiveRecord::Base
     scope :sort_by_phase_name_asc, -> { order("CASE register_phase WHEN 'Beta' THEN 2 WHEN 'Alpha' THEN 3 WHEN 'Discovery' THEN 4 WHEN 'Backlog' THEN 5 END") }
     scope :sort_by_phase_name_desc, -> { order("CASE register_phase WHEN 'Backlog' THEN 1 WHEN 'Discovery' THEN 2 WHEN 'Alpha' THEN 3 WHEN 'Beta' THEN 4 END") }
 
+    def records?
+      Record.where(spina_register_id: id).exists?
```
danmitchell-: Should be `register_id` now

app/controllers/pages_controller.rb
```diff
@@ -22,6 +22,6 @@ def case_study; end
 private
 
   def get_ready_to_use_registers
-    @beta_registers = Spina::Register.where(register_phase: 'Beta')
+    @beta_registers = Spina::Register.where(register_phase: 'Beta').select(&:records?)
```
danmitchell-: Should be `Register` now instead of `Spina::Register`

PR #150, opened 2018-02-01
pipeline page should not 500 if records not populated in DB

gidsg:
### Context
Previously the pipeline page would 500 on `get_register_definition` if any of the non-backlog registers were not populated in the database

Pipeline page should only show registers where data is populated PLUS registers in backlog

### Guidance to review
Pipeline page should still load (e.g. if Charity has not been populated)

### DIFF
--- a/app/controllers/registers_controller.rb
+++ b/app/controllers/registers_controller.rb
@@ -6,17 +6,16 @@ class RegistersController < ApplicationController
   layout 'layouts/default/application'
 
   def index
-    @search = Register.ransack(params[:q])
+    @search = Register.available.ransack(params[:q])
 
-    @registers = if params[:phase] == 'ready to use'
+    @registers = case params[:phase]
+                 when 'ready to use'
                    @search.result.where(register_phase: 'Beta').sort_by_phase_name_asc.by_name
-                 elsif params[:phase] == 'in progress'
+                 when 'in progress'
                    @search.result.where.not(register_phase: 'Beta').sort_by_phase_name_asc.by_name
                  else
                    @search.result.sort_by_phase_name_asc.by_name
                  end
-
-    @current_phases = Register::CURRENT_PHASES
   end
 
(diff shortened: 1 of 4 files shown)


app/views/registers/index.html.haml
```diff
@@ -87,7 +87,7 @@
                 = sort_link(@search, :register_authority, "Managed by", default_order: :asc, hide_indicator: true)
               %th{role: "columnheader", scope: "col"}
           %tbody
-            - @registers.each_with_index do |register, index|
+            - @registers.select(&:available?).each_with_index do |register, index|
```
danmitchell-: Can we move the method into the controller?

`@registers = Register.something.select(&:available?)`

app/controllers/registers_controller.rb
```diff
@@ -6,15 +6,11 @@ class RegistersController < ApplicationController
...
-    @registers = if params[:phase] == 'ready to use'
-                   @search.result.where(register_phase: 'Beta').sort_by_phase_name_asc.by_name
-                 elsif params[:phase] == 'in progress'
-                   @search.result.where.not(register_phase: 'Beta').sort_by_phase_name_asc.by_name
-                 else
-                   @search.result.sort_by_phase_name_asc.by_name
-                 end
+    @search = Register.available.ransack(params[:q])
+
+    @registers = @search.result.sort_by_phase_name_asc.by_name
```
gidsg: can we make these variants a case statement?

app/views/registers/index.html.haml
```diff
@@ -48,7 +48,7 @@
   - else
     .grid-row
       .column-full
-        %h1 Registers pipeline
+        %h1.heading-xlarge Registers pipeline
```
gidsg: did you intend this change to go in there?

danmitchell-: Yeah. #cleanasyougo

PR #154, opened 2018-02-05
Refactoring methods [WIP]

danmitchell-:


app/models/register.rb
```diff
@@ -18,6 +18,14 @@ def records?
     Record.where(register_id: id).exists?
   end
 
+  def last_updated
+    Record.select('timestamp')
+      .where(register_id: self.id, entry_type: 'user')
```
gidsg: Would we consider a `system` update to bump the last updated time? e.g. change of custodian or field description? CC @michaelabenyohai

PR #162, opened 2018-02-08
 catch InvalidRegisterError thrown by client library and delete records and entries

gidsg:
### Context
The ruby client library, as of v0.8.0, has the ability to detect when the data held in memory (whether in memory, postgres etc) conflicts with the data that is held by the register. Specifically, when a register has been reloaded or when the downloaded RSF file does not yield the correct root hash that the register holds, it will raise an `InvalidRegisterError`. Currently the frontend does not have any special handling for this exception which means that registers in this state will never be updated in the frontend (for example when alpha registers are reloaded)

catch `InvalidRegisterError` exception from client library, delete records and entries and throw a new exception. We then rely on the queue retry system to re-populate the register from scratch.

### Guidance to review
When a register is has invalid data its records and entries should be deleted from the frontend db. On a subsequent run, the data should be re-populated.

### DIFF
--- a/app/jobs/populate_register_data_in_db_job.rb
+++ b/app/jobs/populate_register_data_in_db_job.rb
@@ -13,8 +13,10 @@ class PopulateRegisterDataInDbJob < ApplicationJob
       register_client.refresh_data
     rescue InvalidRegisterError
       # If register data is invalid we want to delete existing entries and records to force a full reload
-      Record.where(register_id: register.id).destroy_all
-      Entry.where(register_id: register.id).destroy_all
+      ActiveRecord::Base.transaction do
+        Record.where(register_id: register.id).destroy_all
+        Entry.where(register_id: register.id).destroy_all
+      end
       raise Exceptions::FrontendInvalidRegisterError
     end
 
--- a/app/jobs/populate_register_data_in_db_job.rb
+++ b/app/jobs/populate_register_data_in_db_job.rb
@@ -11,13 +11,13 @@ class PopulateRegisterDataInDbJob < ApplicationJob
     # Initialize client and download / update data
       register_client = @@registers_client.get_register(register.name.parameterize, register.register_phase.downcase, PostgresDataStore.new(register))
       register_client.refresh_data
-    rescue InvalidRegisterError
+    rescue InvalidRegisterError => e
       # If register data is invalid we want to delete existing entries and records to force a full reload
       ActiveRecord::Base.transaction do
         Record.where(register_id: register.id).destroy_all
         Entry.where(register_id: register.id).destroy_all
       end
-      raise Exceptions::FrontendInvalidRegisterError
+      raise Exceptions::FrontendInvalidRegisterError, e
     end
 
     metadata_records = Record.where(register_id: register.id, entry_type: :system)
--- a/app/jobs/populate_register_data_in_db_job.rb
+++ b/app/jobs/populate_register_data_in_db_job.rb
@@ -1,16 +1,21 @@
 class PopulateRegisterDataInDbJob < ApplicationJob
   queue_as :default
 
+  module Exceptions
+    class FrontendInvalidRegisterError < StandardError; end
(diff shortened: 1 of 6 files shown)


app/jobs/populate_register_data_in_db_job.rb
```diff
@@ -1,6 +1,30 @@
...
+  end
+
+  module RegisterClientExtensions
+    include Exceptions
+    def refresh_data
+      begin
+        super
+      rescue InvalidRegisterError
+        # If register data is invalid we want to delete existing entries and records to force a full reload
+        register_id = @data_store.instance_variable_get(:@register).id
```
karlbaker02: I think this is acceptable in tests but perhaps not in production code. Can the `InvalidRegisterException` be caught by wrapping the call to `refreshData` in a try-catch block?

gidsg: No, that is not sufficient because `refresh_data` is also called by the client library itself when `RegisterClient` is initialised in: https://github.com/openregister/registers-ruby-client/blob/2674bb6292d9d6b582c92ac62722bfed8e7a6982/lib/register_client.rb#L12 and this is the first exception thrown.
I agree there should be a better way to access the register ID, since it is  passed to the `perform` method but not available at the class level to be injected here.

arnau: @gidsg I'm not following your explanation.

app/jobs/populate_register_data_in_db_job.rb
```diff
@@ -1,11 +1,22 @@
...
+    begin
     # Initialize client and download / update data
-    register_client = @@registers_client.get_register(register.name.parameterize, register.register_phase.downcase, PostgresDataStore.new(register))
-    register_client.refresh_data
+      register_client = @@registers_client.get_register(register.name.parameterize, register.register_phase.downcase, PostgresDataStore.new(register))
+      register_client.refresh_data
+    rescue InvalidRegisterError
+      # If register data is invalid we want to delete existing entries and records to force a full reload
+      Record.where(register_id: register.id).destroy_all
+      Entry.where(register_id: register.id).destroy_all
```
arnau: Do we want to control that both `destroy_all` are successful? Or, what happens if the first succeeds but the second fails? Is it a tolerable inconsistency?

danmitchell-: id personally use `find` here instead of `where`

gidsg: @danmitchell- I don't _think_ `find` works for this case, as we don't want to delete one item by ID. We want to destroy all matching results

spec/jobs/handle_invalid_register_spec.rb
```diff
@@ -0,0 +1,49 @@
+require 'rails_helper'
+
+RSpec.configure do |config|
```
danmitchell-: We should move this into a global file as every spec is using it.

app/jobs/populate_register_data_in_db_job.rb
```diff
@@ -1,11 +1,24 @@
...
+    begin
     # Initialize client and download / update data
-    register_client = @@registers_client.get_register(register.name.parameterize, register.register_phase.downcase, PostgresDataStore.new(register))
-    register_client.refresh_data
+      register_client = @@registers_client.get_register(register.name.parameterize, register.register_phase.downcase, PostgresDataStore.new(register))
+      register_client.refresh_data
+    rescue InvalidRegisterError
+      # If register data is invalid we want to delete existing entries and records to force a full reload
+      ActiveRecord::Base.transaction do
+        Record.where(register_id: register.id).destroy_all
```
danmitchell-: update to `Record.find(register.id).destroy_all` ?

gidsg: This does not work as `find` selects one record and `destroy_all` works on multiple.

PR #164, opened 2018-02-08
Add user research manifest file

danmitchell-:
### Context
We want to have a staging and user research environment

Add manifest for staging and research environment on paas

### Guidance to review
`cf target -s target sandbox`
`cf push manifest-#{ENVIRONMENT}.yml`

### DIFF
--- a/manifest-production.yml
+++ b/manifest-production.yml
@@ -1,7 +1,10 @@
 ---
 
 inherit: manifest-base.yml
-
+domain: registers-trial.service.gov.uk
+no-hostname: true
+health-check-type: http
+health-check-http-endpoint: /health_check/standard
 env:
   UA_TRACKING_ID: 'UA-86101042-5'
   services:
--- a/manifest-production.yml
+++ b/manifest-production.yml
@@ -1,0 +1,9 @@
+---
+
+inherit: manifest-base.yml
+
+env:
+  UA_TRACKING_ID: 'UA-86101042-5'
+  services:
+    # postgres database
+    - registers-frontend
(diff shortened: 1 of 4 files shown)


gidsg:
I wonder if the research environment only needs 1 instance of `registers-frontend` as we could tolerate downtime during a deployment?
Also, do we definitely want a `staging` environment. I'm not opposed for ad-hoc testing but think it worth asking the question.

michaelabenyohai:
I agree that I'm not sure we need a staging environment.

So we have thoughts around which PaaS space we will put the research environment in? I would guess not prod.

danmitchell-:
I was going to put research in `sandbox` @michaelabenyohai, do you agree?

michaelabenyohai:
Yeah I think `staging` space makes sense. 

Just had a thought about whether we need the research environment to have it's own database, schedulers and queues? Maybe it's worth only deploying the frontend app against the same database for now. If there is a later requirement for research to be able to add their own registers or register data we can update the manifest then?

danmitchell-:
@michaelabenyohai personally I would have a separate database for the research environment so they can play with that as much as they like without having any effect to production. I don't think we can share schedulers and queues because the app would be in a different `space`.

michaelabenyohai:
Good point, if we share the database we would have to put this in prod environment. Let's go with creating all the apps then.

michaelabenyohai:
I'm happy with this other than potentially addressing @gidsg comment about the UR environment only requiring 1 instance - what do you think @danmitchell- ?

manifest-production.yml
```diff
@@ -0,0 +1,12 @@
+---
+
+inherit: manifest-base.yml
+domain: registers-trial.service.gov.uk
```
michaelabenyohai: Is this section specific to the `registers-frontend` application? The inherited `manifest-base.yml` contains 3 applications so does this definitely override the correct one?

PR #169, opened 2018-02-12
move client manager class var into wrapper class

gidsg:
### Context
Previously on startup we had the Ruby warning: 
```
config/initializers/registers_client.rb:1: warning: class variable access from toplevel
```

Wrap the access to the class var in a class to stop this warning, similar to the pattern described in: 
http://ruby-doc.org/core-2.5.0/Module.html#method-i-class_variable_set

### Guidance to review
Warning should no longer show on startup.


gidsg:
@arnau agreed this is just a first step. I will make a new card for next steps.

PR #171, opened 2018-02-12
change gov orgs to db lookup

gidsg:
### Context
Previously we used the client library to retrieve the list of gov orgs for the CMS

Retrieve gov orgs from the DB. This is part of the separation of concerns whereby the population job should be the only thing using the client library.

### Guidance to review
Check Authority select box in CMS `admin/registers/new` is still populated.

### DIFF
--- a/app/controllers/spina/admin/registers_controller.rb
+++ b/app/controllers/spina/admin/registers_controller.rb
@@ -60,8 +60,9 @@ module Spina
       end
 
       def set_government_organisations
-        register_data = RegistersClientWrapper.registers_client.get_register('government-organisation', 'beta')
-        @government_organisations = register_data.get_records
+        government_organisation_register = Register.find_by(slug: 'government-organisation')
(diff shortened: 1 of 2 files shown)


app/controllers/spina/admin/registers_controller.rb
```diff
@@ -60,8 +60,9 @@ def set_breadcrumb
       end
 
       def set_government_organisations
-        register_data = RegistersClientWrapper.registers_client.get_register('government-organisation', 'beta')
-        @government_organisations = register_data.get_records
+        government_organisation_register = Register.find_by(slug: 'government-organisation')
+        government_organisation_records = Record.where(register_id: government_organisation_register.id, entry_type: 'user')
```
danmitchell-: Can we just call the records from the register? i.e. 
`government_organisation_records  = Register.find_by(slug: 'government-organisation').records.where(entry_type: "user")`

PR #173, opened 2018-02-13
Ui updates

danmitchell-:
### Context
UI review from @conordelahunty 

As per Trello https://trello.com/c/zFDKvgzF/1911-fix-style-bugs-before-deploying-new-frontend-everywhere

### Guidance to review
@conordelahunty 👍

### DIFF
--- a/app/jobs/populate_register_data_in_db_job.rb
+++ b/app/jobs/populate_register_data_in_db_job.rb
@@ -10,6 +10,7 @@ class PopulateRegisterDataInDbJob < ApplicationJob
     begin
     # Initialize client and download / update data
       register_client = RegistersClientWrapper.registers_client.get_register(register.name.parameterize, register.register_phase.downcase, PostgresDataStore.new(register))
+      register_url = register_client.instance_variable_get(:@register_url)
       register_client.refresh_data
     rescue InvalidRegisterError => e
     # If register data is invalid we want to delete existing entries and records to force a full reload
@@ -23,6 +24,7 @@ class PopulateRegisterDataInDbJob < ApplicationJob
     metadata_records = Record.where(register_id: register.id, entry_type: :system)
     fields = metadata_records.select { |record| record.key.start_with?("field:") }
     register.fields = fields.map { |field| field.data['field'] }.join(',')
+    register.url = register_url
     register.save
   end
 end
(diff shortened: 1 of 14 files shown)


app/views/registers/show.html.haml
```diff
@@ -65,14 +65,12 @@
...
-          %span.summary Download a copy of this register
-        #details-content-2.panel.panel-border-narrow{"aria-hidden" => "true"}
-          %p Download this data as: #{link_to 'TSV', "https://#{@register.name.parameterize}.register.gov.uk/records.tsv"}, #{link_to 'CSV', "https://#{@register.name.parameterize}.register.gov.uk/records.csv"} or as a #{link_to 'Spreadsheet', "https://#{@register.name.parameterize}.register.gov.uk/records.xlsx"}.
-          %p Preview a sample as #{link_to 'JSON', "https://#{@register.name.parameterize}.register.gov.uk/preview-records/json", target: :blank}, #{link_to 'YAML', "https://#{@register.name.parameterize}.register.gov.uk/preview-records/yaml", target: :blank} or #{link_to 'TTL', "https://#{@register.name.parameterize}.register.gov.uk/preview-records/turtle", target: :blank}.
+        %li
+          - if @total_record_count > 5000
+            download the first 5,000 records
+          - else
+            download a copy of this register
+          as a #{link_to 'CSV', "https://#{@register.name.parameterize}.register.gov.uk/records.csv"} and keep your copy up-to-date
```
gidsg: hardcoding to `register.gov.uk` won't work for non ready to use registers. You need to work out the url based on the phase.

gidsg: You want to add `?page-size=5000` to this URL.

danmitchell-: @gidsg I agree however I was under the impression that discovery registers now on cloudapps were "throw away" and not for the general public to view so we wouldnt show the `view register` link for those on the collection page? cc @elliecraven

elliecraven:
confirm - discovery registers aren't intended for general public view.

app/views/registers/show.html.haml
gidsg: We still need a solution for alpha registers, so I would suggest reading the value from the client library, as that is the 'source of truth'.

app/jobs/populate_register_data_in_db_job.rb
```diff
@@ -10,6 +10,7 @@ def perform(register)
     begin
     # Initialize client and download / update data
       register_client = RegistersClientWrapper.registers_client.get_register(register.name.parameterize, register.register_phase.downcase, PostgresDataStore.new(register))
+      register_url = register_client.instance_variable_get(:@register_url)
```
gidsg: I think this should be a public method on the client library but that needs an upstream change, so this is OK for now.

danmitchell-: 👍

PR #175, opened 2018-02-14
Add and enable prometheus metrics endpoint

fjavierm:
### Context
Add and enable Prometheus metrics endpoint to be able to have consistent information about the application.

Addition of the instrumentation gem

### Guidance to review
https://github.com/alphagov/gds_metrics_ruby


gidsg:
@michaelabenyohai my understanding is that this won't do anything yet until we have added some additional configuration in the app, but I think it is safe to merge as-is.

PR #177, opened 2018-02-14
Remove GDS Transport font

danmitchell-:
:(


app/assets/stylesheets/application.scss
```diff
@@ -1,5 +1,9 @@
 @charset "utf-8";
 
+// Cannot use GDS Transport font on a non GOV.UK domain due to licensing
```
gidsg: Is there any way to add this conditionally only on the trial domain?

danmitchell-: I can check for the host and conditionally add a class name to the body but I don't think its worth the overhead/code bloat....

PR #188, opened 2018-02-19
add Links From

gidsg:
Add Links From metadata based on the [definition in ORJ](https://github.com/openregister/openregister-java/blob/9df5498a67568869ab9f5de0ed54d51e466e94b2/src/main/java/uk/gov/register/service/RegisterLinkService.java#L37): 
screen shot 2018-02-19 at 15 52 31

### Guidance to review
To test locally you will need to add a linked from register e.g. `school-type-eng`

### DIFF
--- a/app/models/register.rb
+++ b/app/models/register.rb
@@ -45,6 +45,7 @@ class Register < ApplicationRecord
   def links_from
     Register.where.not(id: id)
     .joins(:records)
+    .where(records: { entry_type: 'system' })
     .where("data->>'register' = ?", slug)
   end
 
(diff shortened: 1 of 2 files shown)


app/models/register.rb
```diff
@@ -42,6 +42,11 @@ def links_to
     Register.where(slug: linked_register_names)
   end
 
+  def links_from
+    Register.where("? = ANY(string_to_array(fields,','))", slug)
```
gidsg: we have to do this SQL magic because we suboptimally store the fields as `text` rather than an `array` 

we should make a backlog card to fix this.

gidsg: ...actually I think to do this properly we should base it solely on the 'register' property of fields, not the field name alone.

PR #201, opened 2018-02-21
keep filter if no results found

### LINKED ISSUE #191: Filter options auto-collapse when no results are found
### What did you do? (steps to reproduce)

Go to any register page on the frontend that has no archived records (for example Local Authority type). Expand the filter options and click on the `Archived records` radio button.

### What happened? (actual results)

When we see that there are no results, the filter group auto-collapses.

### What should have happened? (expected results)

I expected the filter group to stay visible so that I can easily change my filter selection, without having to first expand the filter options.

gidsg:
### Context
Fixes #191

Maintain filter options if no results found.

### Guidance to review
See steps to reproduce in #191

### DIFF
--- a/app/views/registers/show.html.haml
+++ b/app/views/registers/show.html.haml
@@ -89,8 +89,8 @@
                 = search_field_tag 'q', nil, class: 'search-input', placeholder: 'Search', value: params[:q]
                 = submit_tag 'Search', name: nil, class: 'search-submit'
             .column-full
-              %details{role: "group", open: (params[:status].present? && @records.present? ? 'open' : nil)}
-                %summary{"aria-controls" => "details-content-0", "aria-expanded" => "#{params[:status].present? && @records.present? ? 'true' : 'false'}", role: "button"}
+              %details{role: "group", open: (params[:status].present? ? 'open' : nil)}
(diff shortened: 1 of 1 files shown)


app/views/registers/show.html.haml
```diff
@@ -89,8 +89,8 @@
                 = search_field_tag 'q', nil, class: 'search-input', placeholder: 'Search', value: params[:q]
                 = submit_tag 'Search', name: nil, class: 'search-submit'
             .column-full
-              %details{role: "group", open: (params[:status].present? && @records.present? ? 'open' : nil)}
```
danmitchell-: I wonder if we should check if `params[:q]` has a value as opening the filter does seem pointless if we are showing the same data anyway.

gidsg: The case here is when you have selected a `params[:status]` e.g. Archived but there are no results. So in that case it is useful to see the filter even if there is no `params[:q]

danmitchell-: Ah okay. I tested this by searching and empty query which just reloaded the page and opened the filter section which seemed off but I see what you're saying 👌

gidsg: Erm, this was the behaviour already in `beta-assessment`. So is unrelated to my PR. It's becuase the status param is set by default in the form even though it is not visible.

PR #207, opened 2018-02-26
Add new suggest register flow

danmitchell-:
### Context
We will be doing a lot of product site changes, this new form will be part of that work so we're doing this in preparation

Add new "suggest register" flow ready for when the product site UI has been updated.

### Guidance to review
`/suggest-register`

### DIFF
--- a/app/controllers/suggest_registers_controller.rb
+++ b/app/controllers/suggest_registers_controller.rb
@@ -1,6 +1,6 @@
 class SuggestRegistersController < ApplicationController
   def merge_from_params
-    session[:suggest_register_params] ? session[:suggest_register_params].merge(params.permit(:email, :message, :subject)).except('step') : {}
+    session[:suggest_register_params] ? session[:suggest_register_params].except('step') : {}
   end
 
   def index
--- a/app/controllers/suggest_registers_controller.rb
+++ b/app/controllers/suggest_registers_controller.rb
@@ -9,22 +9,18 @@ class SuggestRegistersController < ApplicationController
     @suggest_register = @wizard.object
 
     if @wizard.save
-      debugger
       redirect_to root_path
     else
       render :index
     end
   end
 
-  def thanks
+  def thanks; end
 
-  end
-
-  private
+private
 
   def suggest_register_params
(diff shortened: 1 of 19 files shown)


app/controllers/suggest_registers_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  def create
+    @wizard = ModelWizard.new(SuggestRegister.new(merge_from_params), session, params, suggest_register_params).continue
+    @suggest_register = @wizard.object
+
+    if @wizard.save
+      @zendesk_service = ZendeskFeedback.new
+      @response = @zendesk_service.send_feedback(suggest_register_params)
+      session[:suggest_register_params] = nil
+      redirect_to suggest_register_complete_path
```
karlbaker02: Where is `suggest_register_complete_path` defined?

app/views/suggest_registers/steps/_page1.html.haml
```diff
@@ -0,0 +1,15 @@
+%h1.heading-large Why are you suggesting in this register?
```
karlbaker02: _Why are you suggesting this register?_

karlbaker02:
Should the lines above and below the `Change` link extend all the way to the end of the link, or past the link? Some funky looking behaviour going on here. Note also that there is `0` padding for the first `cya-answer`.

screen shot 2018-02-26 at 15 27 53

karlbaker02:
Is it possible when you click on the `Change` link on the _Check your answers_ page to only go to the page for that particular change and then straight back to the _Check your answers_ page?

An example being that if you change your email address, because this is the last step in the pipeline, you end up on the _Check your answers_ page immediately after submitting. 

If instead you need to change the _Why are you interested_ answer, you do this and then go to the _Register title_ page, then to the _Email address_ page before finally ending back up at the summary.

danmitchell-:
@karlbaker02 going through the flow after changing one answer is the pattern for this "smart answer" journey.

karlbaker02:
I see. Is that a GDS pattern?

PR #211, opened 2018-02-26
Add new request register flow

danmitchell-:
### Context
We will be doing a lot of product site changes, this new form will be part of that work so we're doing this in preparation

Add new "express interest in register" flow ready for when the product site UI has been updated.

### Guidance to review
`/request-register?register_title=charity`


app/controllers/suggest_registers_controller.rb
```diff
@@ -0,0 +1,33 @@
+class SuggestRegistersController < ApplicationController
```
karlbaker02: Are you meaning to include the `SuggestRegistersController` again in this PR?

app/models/suggest_register.rb
```diff
@@ -0,0 +1,11 @@
+class SuggestRegister < ApplicationRecord
```
karlbaker02: Same again for this.

PR #212, opened 2018-02-27
Use page_entries_info from Kaminari to replace custom code

danmitchell-:
### Context
Removing some handwritten code and take advantage of built in features from gems such as Kaminari

Remove custom pagination code in the `recover_records` methods

### Guidance to review
Open a register and click "load more"

### DIFF
--- a/app/controllers/registers_controller.rb
+++ b/app/controllers/registers_controller.rb
@@ -35,6 +35,7 @@ private
     sort_by = params[:sort_by] ||= default_sort_by.call
     sort_direction = params[:sort_direction] ||= 'asc'
     query = @register.records.where(entry_type: 'user')
+    @total_record_count = query.count
 
     query = case status
             when 'archived'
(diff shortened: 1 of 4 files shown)


app/controllers/registers_controller.rb
```diff
@@ -63,12 +62,7 @@ def recover_records(fields, params, page_size = 100)
...
       query = query.order("data->> '#{sort_by}' #{sort_direction.upcase} nulls last")
     end
-    @result_count = query.count
-    @offset = page_size * (page - 1) + 1
-    @offset_end = [@result_count, page_size * page].min
 
-    @total_pages = (@result_count / 100) + ((@result_count % 100).zero? ? 0 : 1)
-
-    query.page(page).per(page_size).without_count
+    query.page(params[:page]).per(100)
```
karlbaker02: This probably isn't a regression on what we were doing before, but do I understand correctly that we'll run the `where` statement on the query first, get back everything that matches that filter, and then only take the `params[:page]` page of 100 results?

danmitchell-: I think so, yes. We have stories in the icebox to refactor the two parts of this method anyway. Search and Sorting.

PR #215, opened 2018-02-27
store fields as array in DB

gidsg:
### Context
Previously we stored fields in the DB as a comma separated string in the order from the RSF (which may not match the register definition). This meant we were frequently turning it back into an array in the app logic and also performing additional queries to get the order.

Store fields in the DB as an array ordered based on the register definition. 

### Guidance to review
This is a prerequisite to future refactoring once this migration is applied and the data is populated we can deprecate the previous `fields` column in the DB.


karlbaker02:
It might still be useful to have a check in case the register definition has been updated with new fields and/or re-ordering of the existing fields.

gidsg:
@karlbaker02 this will update on every re-population. So it should always be the latest version. CC @michaelabenyohai

danmitchell-:
Nice!

PR #216, opened 2018-02-27
Refactor entries pagination to use Kaminari

danmitchell-:
Use Kaminari method to control pagination

### Guidance to review
Visit an "updates" page, search and "load more"

### DIFF
--- a/app/controllers/entries_controller.rb
+++ b/app/controllers/entries_controller.rb
@@ -22,29 +22,24 @@ class EntriesController < ApplicationController
       { current_record: current_record, previous_record: previous_record, updated_fields: changed_fields, key: current_record.key }
     end
 
-    @entries_with_items = Kaminari::paginate_array(@entries_with_items, total_count: @result_count).page(@current_page).per(100)
+    @entries_with_items = Kaminari::paginate_array(@entries_with_items).page(params[:page]).per(100)
   end
 
 private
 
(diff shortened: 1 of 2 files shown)


app/controllers/entries_controller.rb
```diff
@@ -22,29 +22,24 @@ def index
...
       fields.split(',').each { |field| partial += " data->> '#{field}' ilike ? or" }
       partial = partial[1, partial.length - 3] # Remove beginning space and end 'or'
 
       operation_params.push(partial)
-      fields.split(',').count.times { operation_params.push("%#{search_term}%") }
+      fields.split(',').count.times { operation_params.push("%#{params[:q]}%") }
 
-      query = Entry.where(register_id: register_id, entry_type: 'user').where(operation_params).order(:entry_number).reverse_order.limit(page_size).offset(page_size * (page - 1))
-      count_query = Entry.where(register_id: register_id, entry_type: 'user').where(operation_params)
+      query = Entry.where(register_id: register_id, entry_type: 'user').where(operation_params).order(:entry_number).reverse_order
```
gidsg: This query does not have a limit so is very expensive for large registers.

danmitchell-: Ah. Good spot. I'll add `.limit(100)` back

gidsg: will you not also need the offset as before to provide the correct results?

PR #218, opened 2018-03-01
Use fields array everywhere

gidsg:
### Context
Follow up to #215 

- use `fields_array` everywhere
- move search used by `Record` and `Entry` into Concern

### Guidance to review
This is a refactor everything should work the same.

### DIFF
--- a/app/controllers/entries_controller.rb
+++ b/app/controllers/entries_controller.rb
@@ -29,12 +29,13 @@ private
     search_term = params[:q]
     page = params.fetch(:page) { 1 }.to_i
     user_entries = Entry.where(register_id: register_id, entry_type: 'user')
+    query = user_entries.with_limit(page, page_size)
+
     if search_term.present?
       search_query = user_entries.search_for(fields, search_term)
-      query = search_query.order(:entry_number).limit(page_size).offset(page_size * (page - 1))
+      query = search_query.with_limit(page, page_size)
       count_query = search_query
     else
-      query = user_entries.order(:entry_number).reverse_order.limit(100).offset(100 * (page - 1))
       count_query = user_entries
     end
 
--- a/app/controllers/entries_controller.rb
+++ b/app/controllers/entries_controller.rb
@@ -28,13 +28,14 @@ private
   def recover_entries_history(register_id, fields, params, page_size = 100)
     search_term = params[:q]
     page = params.fetch(:page) { 1 }.to_i
-
+    user_entries = Entry.where(register_id: register_id, entry_type: 'user')
     if search_term.present?
-      query = Entry.where(register_id: register_id, entry_type: 'user').search_for(fields, search_term).order(:entry_number).limit(page_size).offset(page_size * (page - 1))
-      count_query = Entry.where(register_id: register_id, entry_type: 'user').search_for(fields, search_term)
+      search_query = user_entries.search_for(fields, search_term)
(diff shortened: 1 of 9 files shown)


danmitchell-:
Really nice clean up 👍 🥇

app/models/entry.rb
```diff
@@ -1,4 +1,5 @@
 class Entry < ApplicationRecord
   has_one :record
   belongs_to :register
+  include SearchScope
```
danmitchell-: Move to line 2 to match `records.rb`?

app/controllers/entries_controller.rb
```diff
@@ -5,8 +5,8 @@ class EntriesController < ApplicationController
 
   def index
     @register = Register.find_by_slug!(params[:register_id])
-    entries = recover_entries_history(@register.id, @register.fields, params)
-    fields = @register.register_fields.map { |field| field['field'] }
+    fields = @register.fields_array
```
danmitchell-: I don't think we need to create this as variable. Might read easier in the method to just specify the array directly`@register.fields_array`

gidsg: since we are using this multiple times, I assume this ensures we are only doing the DB lookup once, but I am not sure how this works in Ruby/Rails internals.

arnau: mmm, I only see one call to `#fields_array` in this method. I would expect this field to be fetched from the database every time you instantiate `@register` rather than when you call this method. Am I missing something?

app/controllers/entries_controller.rb
```diff
@@ -31,16 +31,8 @@ def recover_entries_history(register_id, fields, params, page_size = 100)
     page = (params[:page] ||= 1).to_i
```
arnau: this seems to be better as `page = (params[:page] || 1).to_i` or more my taste `page = params.fetch(:page) { 1 }.to_i` so you are expressing your intention that you want to access `:page` and want to use `1` as the default value instead of `nil`.

app/controllers/entries_controller.rb
```diff
@@ -31,16 +31,8 @@ def recover_entries_history(register_id, fields, params, page_size = 100)
...
-      operation_params = []
-      fields.split(',').each { |field| partial += " data->> '#{field}' ilike ? or" }
-      partial = partial[1, partial.length - 3] # Remove beginning space and end 'or'
-
-      operation_params.push(partial)
-      fields.split(',').count.times { operation_params.push("%#{search_term}%") }
-
-      query = Entry.where(register_id: register_id, entry_type: 'user').where(operation_params).order(:entry_number).reverse_order.limit(page_size).offset(page_size * (page - 1))
-      count_query = Entry.where(register_id: register_id, entry_type: 'user').where(operation_params)
+      query = Entry.where(register_id: register_id, entry_type: 'user').search_for(fields, search_term).order(:entry_number).limit(page_size).offset(page_size * (page - 1))
```
arnau: We could benefit from some dryness here by doing

```ruby
query = Entry.where(register_id: register_id, entry_type: 'user')
```

Also, it seems that the linter is happy with these long lines but, do you find them readable? I'll admit that beyond 80 chars things start to be hard for me to follow, more than 100 it feels too much.

app/controllers/entries_controller.rb
gidsg: Yes, I think you are right, I will amend as-per @danmitchell- suggestion

PR #228, opened 2018-03-08
Redesign collection page

danmitchell-:
https://projects.invisionapp.com/share/M8G0D1XX96F#/screens/283459566

### Guidance to review
`/registers` and then `/upcoming-registers`


gidsg:
we will want to redirect `/registers?phase=in+progress` to the new `registers-in-progress` I am not a big fan of changing URLs but I guess you need to do it in this case.

PR #234, opened 2018-03-09
do not deploy queue and scheduler on beta-assessment branch

gidsg:
do not deploy queue and scheduler on beta-assessment branch as it was failing since they are no longer defined in the maniferst

### Guidance to review
Deploy should be successful from travis


danmitchell-:
Do we want to remove `SINGLE_INSTANCE_APPS=(registers-frontend-queue registers-frontend-scheduler)` as well?

PR #236, opened 2018-03-12
Add tabs to registers show

danmitchell-:
Split `registers#show` into tabs

### Guidance to review
Visit `registers#show` and click through each tab


gidsg:
These tabs don't seem to be keyboard tab-able?

gidsg:
Feels kinda weird that *Details* is the same size/weight as the stuff below but acts like a higher heading? CC @conordelahunty 
screen shot 2018-03-12 at 17 25 43

danmitchell-:
You are right @gidsg. The aim of this work is to just get the structure in and then there will be a full design review with Connor. The work is behind closed doors so we can afford to be abit informal with it.

PR #238, opened 2018-03-13
BAU: Bug fix to support controller

AmritSidhu:
I found this bug while trying to replicate the support form for the Connecting to verify team. 
When user submits an invalid form, the support controller caused the fields to change.
Example: The title on 'Report a problem' page changes to 'Ask a question or give feedback' upon invalid submission.

### DIFF
--- a/app/controllers/support_controller.rb
+++ b/app/controllers/support_controller.rb
@@ -6,7 +6,7 @@ class SupportController < ApplicationController
   def index; end
 
   def select_support
-    if params[:subject] == "problem"
+    if params[:subject] == '[Problem]'
       redirect_to support_problem_path
     else
       redirect_to support_question_path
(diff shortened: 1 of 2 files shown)


gidsg:
Hi @AmritSidhu thanks for your contribution, there are some rubocop errors preventing the build from passing, please can you fix these e.g. `bundle exec govuk-lint-ruby -a app config db lib spec `

AmritSidhu:
All done, removed the new lines causing the errors.

app/controllers/support_controller.rb
```diff
@@ -30,7 +30,7 @@ def create
       redirect_to support_thanks_path
     else
       flash[:errors] = @support.errors
-      if params[:subject] == "problem"
+      if params[:support][:subject] == "[Problem]"
```
gidsg: very minor but you don't need double quotes here as you are not using string interpolation.

AmritSidhu: No problem, Ive changed them to single quotes along with "problem" to match "[Problem]"

PR #245, opened 2018-04-04
Add http auth to post method

danmitchell-:
Post to self service app with auth credentials

### Guidance to review
`/api_key/new`, add email and submit.

### DIFF
--- a/app/controllers/api_users_controller.rb
+++ b/app/controllers/api_users_controller.rb
@@ -27,7 +27,16 @@ private
 
   def post_to_endpoint(user)
     @user = { email: user.email, department: user.department, service: user.service }
-    Net::HTTP.post_form(URI(Rails.configuration.self_service_api_endpoint), @user)
+
+    uri = URI.parse(Rails.configuration.self_service_api_endpoint)
+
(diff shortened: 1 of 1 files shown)


app/controllers/api_users_controller.rb
```diff
@@ -27,7 +27,16 @@ def create
...
     @user = { email: user.email, department: user.department, service: user.service }
-    Net::HTTP.post_form(URI(Rails.configuration.self_service_api_endpoint), @user)
+
+    uri = URI.parse(Rails.configuration.self_service_api_endpoint)
+
+    Net::HTTP.start(uri.host, uri.port) do |http|
+      header = { 'Content-Type': 'application/x-www-form-urlencoded' }
+      request = Net::HTTP::Post.new(uri.request_uri, header)
+      request.basic_auth(ENV['SELF_SERVICE_HTTP_AUTH_USERNAME'], ENV['SELF_SERVICE_HTTP_AUTH_PASSWORD'])
+      request.body = URI.encode_www_form(@user)
```
gidsg: looks like if you use `set_form_data` then that sets the correct header for you.

PR #257, opened 2018-04-11
add error handling in api key post

gidsg:
### Context
Handle case where we get an unexpected HTTP response from registers-selfservice

Catch exceptions thrown by `net/http` and show error message

Note: I am aware this does not handle the case where we get a valid (e.g. 200 response from registers-selfservice but with malformed content, this will be a separate PR.)

## Before
screen shot 2018-04-11 at 10 19 34

## After
screen shot 2018-04-11 at 11 30 30

### Guidance to review
You can test this locally by attempting to submit the api key form whilst `registers-selfservice` app is not running

### DIFF
--- a/app/controllers/api_users_controller.rb
+++ b/app/controllers/api_users_controller.rb
@@ -19,7 +19,7 @@ class ApiUsersController < ApplicationController
         render :show
       else
         flash.alert = 'Something went wrong'
-        logger.error("API Key POST failed with unexpected response code: #{response&.code}")
+        logger.error("API Key POST failed with unexpected response code: #{response.code}") if response&.code
         render :new
       end
     end
@@ -31,7 +31,7 @@ private
     @user = { email: user.email, department: user.department, service: user.service }
     uri = URI.parse(Rails.configuration.self_service_api_endpoint)
     options = {
-      basic_auth: { username: ENV['SELF_SERVICE_HTTP_AUTH_USERNAME'], password: ENV['SELF_SERVICE_HTTP_AUTH_PASSWORD'] },
+      basic_auth: { username: ENV.fetch['SELF_SERVICE_HTTP_AUTH_USERNAME'], password: ENV.fetch['SELF_SERVICE_HTTP_AUTH_PASSWORD'] },
       body: @user
     }
     error_message = 'Something went wrong'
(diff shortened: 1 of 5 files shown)


app/controllers/api_users_controller.rb
```diff
@@ -28,15 +29,19 @@ def create
...
-    http = Net::HTTP.new(uri.host, uri.port)
-    http.use_ssl = (uri.scheme == 'https')
-
-    http.start do |http_start|
-      request = Net::HTTP::Post.new(uri.request_uri)
-      request.basic_auth(ENV['SELF_SERVICE_HTTP_AUTH_USERNAME'], ENV['SELF_SERVICE_HTTP_AUTH_PASSWORD'])
-      request.set_form_data(@user)
-      http_start.request(request)
+    options = {
+      basic_auth: { username: ENV['SELF_SERVICE_HTTP_AUTH_USERNAME'], password: ENV['SELF_SERVICE_HTTP_AUTH_PASSWORD'] },
```
arnau: It would be a good idea to check that both username and password ENV are not `nil`. Ideally in the boundaries of the application. Perhaps a good initial step is to change to `ENV.fetch('SELF_SERVICE_HTTP_AUTH_USERNAME')` so at least the error will be very clear on what is wrong.

arnau: Also, to keep the docker deploy working we should probably have a `docker-compose.yml` on the lines of:

```yaml
version: '3'
services:
  db:
    image: postgres:9.6-alpine
    volumes:
      - data:/var/lib/postgresql/data/pgdata
    environment:
      POSTGRES_PASSWORD:
      POSTGRES_USER: openregister-info
      POSTGRES_DB: openregister-info_development
      PGDATA: /var/lib/postgresql/data/pgdata
  web:
    build: .
    ports:
      - "3000:3000"
    depends_on:
      - db
    environment:
      SELF_SERVICE_HTTP_AUTH_USERNAME: ${SELF_SERVICE_HTTP_AUTH_USERNAME}
      SELF_SERVICE_HTTP_AUTH_PASSWORD: ${SELF_SERVICE_HTTP_AUTH_PASSWORD}

volumes:
  data: {}
```

PR #259, opened 2018-04-13
add uri check to authentication

gidsg:
### Context
Required to make `beta-assessment-v2` safe to merge into `master`

Only apply authentication if URI is `registers-trial.service.gov.uk`

### Guidance to review
https://registers-trial.service.gov.uk should continue to have auth after merge. https://registers.cloudapps.digital should not have auth.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,7 +1,9 @@
 class ApplicationController < Spina::ApplicationController
   include Spina::Frontend
 
-  http_basic_authenticate_with name: ENV['HTTP_AUTH_USERNAME'], password: ENV['HTTP_AUTH_PASSWORD'] if Rails.env.production?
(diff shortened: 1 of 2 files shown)


app/controllers/application_controller.rb
```diff
@@ -1,7 +1,9 @@
 class ApplicationController < Spina::ApplicationController
   include Spina::Frontend
 
-  http_basic_authenticate_with name: ENV['HTTP_AUTH_USERNAME'], password: ENV['HTTP_AUTH_PASSWORD'] if Rails.env.production?
+  is_trial_domain = Rails.env.production? && JSON.parse(ENV.fetch('VCAP_APPLICATION'))['uris'].include?('registers-trial.service.gov.uk')
+
+  http_basic_authenticate_with name: ENV['HTTP_AUTH_USERNAME'], password: ENV['HTTP_AUTH_PASSWORD'] if is_trial_domain
```
arnau: Shouldn't the env variables be moved to the Rails config file?

PR #263, opened 2018-04-23
Remove SpinaCMS

danmitchell-:
### Context
The frontend uses a piece of technology called Spina CMS. It was originally used because the frontend used to allow the team to create/change large pieces of text for the history pages. These history pages no longer exist and Spina CMS is only used to manage the pipeline page of registers. Using Spina CMS is overkill for this and an unnecessary dependency.

Remove Spina CMS

### Guidance to review
App should perform as before


config/routes.rb
```diff
@@ -46,15 +46,14 @@
...
-    namespace :admin do
-      resources :registers, except: [:show]
-    end
+  namespace :admin, path: '/admin' do
+    root to: "registers#index"
+
+    resources :registers, except: [:show]
+    resources :users
+    resources :sessions
+    get "login" => "sessions#new"
```
gidsg: can we make these URLs `/login` and `/logout`? Session is a bit too much implementation detail for me.

gidsg: seems like after failed login you are redirected to `/admin/sessions` can we keep it on `/admin/login`?

app/views/admin/registers/_register.html.haml
```diff
@@ -0,0 +1,6 @@
+%tr
+  %td
+    = link_to register.name, edit_admin_register_path(register)
+  %td= register.register_phase
+  %td
+    = link_to 'Edit', edit_admin_register_path(register)
```
gidsg: maybe just remove this link as it is duplication of link on the register name?

app/views/admin/registers/edit.html.haml
```diff
@@ -0,0 +1,15 @@
+- @page_title = 'Registers | Admin'
```
gidsg: can we include the register name in the title?

app/views/admin/registers/_form.html.haml
```diff
@@ -0,0 +1,28 @@
...
+  = f.text_field :name
+  = f.text_field :description
+  .form-group
+    = f.label :register_phase, class: 'form-label'
+    = f.select :register_phase, Register::CURRENT_PHASES, {}, include_blank: 'Select current phase', class: 'form-control'
+  .form-group
+    = f.label :authority, class: 'form-label'
+    = f.select :authority, @government_organisations.all.collect{ |org| [org.data['name'], org.key] }, {}, class: 'form-control'
+  = f.text_area :contextual_data
+  = f.text_area :related_registers
```
gidsg: It's unclear to me what you are expecting in this textarea. URLs? Markdown?

danmitchell-: They aren't used right now anyway so we can iterate on this at a later date. This entire admin interface will need some UCD input anyway but its low on priorities right now.

app/views/admin/registers/_form.html.haml
```diff
@@ -0,0 +1,28 @@
...
+  .form-group
+    = f.submit 'Save', class: 'button'
+
+  - unless @register.new_record?
+    .pull-right= link_to 'Delete register', admin_register_path(@register), method: :delete, data: { confirm: 'Are you sure you want to delete this record?' }
```
gidsg: should be _"delete this register"_

app/views/admin/registers/_form.html.haml
gidsg: maybe just remove these fields from the UI if we're not using them currently?

spec/support/feature_spec_helpers.rb
```diff
@@ -0,0 +1,12 @@
+module FeatureSpecHelpers
+  def sign_in(user)
+    visit '/admin/signin'
+    fill_in :email, with: user.email
+    fill_in :password, with: 'password123'
+    click_button 'Continue'
```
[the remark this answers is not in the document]
danmitchell-: No, just signing in with the user passed into the method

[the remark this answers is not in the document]
gidsg: I think this isn't a good test then because even if it blew up after clicking continue the test may still pass. You should assert the next URL or some text on the homepage.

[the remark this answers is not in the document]
danmitchell-: This isn't testing the sign in journey itself. We use it in the user and register feature spec to just sign into the admin UI and then go from there.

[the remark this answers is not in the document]
gidsg: Ahhh, OK makes sense 👍

PR #267, opened 2018-04-25
Update service domain

danmitchell-:
We now have `https://www.registers.service.gov.uk/` so we should use that instead of the cloudapps domain.

The gov.uk domain means we can use the transport font.


app/views/pages/case_study.html.haml
```diff
@@ -31,11 +31,11 @@
...
       %p In particular, we found the canonical lists of local authorities in the UK and the list of the many and varied different organisations that make up the government to be very useful. We were able to very quickly integrate our new system with the registers platform via their API and immediately begin using the data.
       %h3.heading-medium GOV.UK Registers are different
-      %p #{link_to 'GOV.UK Registers', 'https://registers.cloudapps.digital', target: :blank} have the following unique properties that make them especially suitable for our needs:
+      %p #{link_to 'GOV.UK Registers', 'https://www.registers.service.gov.uk', target: :blank} have the following unique properties that make them especially suitable for our needs:
```
gidsg: link to `root_path`

PR #272, opened 2018-04-27
send 'is government' response to self-service

gidsg:
### Context
To fulfil suggestion by Dan Gilbert

send 'is government' response in self-service form POST

### Guidance to review
if run with https://github.com/openregister/registers-selfservice/pull/12 'non government user' should be written to department column of sheet.

### DIFF
--- a/app/controllers/api_users_controller.rb
+++ b/app/controllers/api_users_controller.rb
@@ -28,7 +28,10 @@ class ApiUsersController < ApplicationController
 private
 
   def post_to_endpoint(user)
-    @user = { email: user.email, department: user.department, service: user.service }
+    @user = { email: user.email,
+              department: user.department,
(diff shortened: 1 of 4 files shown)


app/controllers/api_users_controller.rb
```diff
@@ -28,7 +28,10 @@ def create
 private
 
   def post_to_endpoint(user)
-    @user = { email: user.email, department: user.department, service: user.service }
+    @user = { email: user.email,
+              department: user.department,
+              service: user.service,
+              is_government: user.is_government == 'yes' }
```
danmitchell-: What is the intention here?

gidsg: we are marshalling the form response `yes` or `no` into a boolean.

PR #274, opened 2018-04-30
Update tech docs links

danmitchell-:
- Update registers-test-docs to docs.registers.service.gov.uk
- Update header links
- Update footer links

### Guidance to review
App should perform as before

# After
screen shot 2018-04-30 at 16 22 38
screen shot 2018-04-30 at 16 22 35

### DIFF
--- a/app/views/pages/using_registers.html.haml
+++ b/app/views/pages/using_registers.html.haml
@@ -39,6 +39,6 @@
         %li download either the whole or part of a register
         %li find a single record, entry or item
 
-      %p The #{link_to 'technical documentation', 'https://docs.registers.service.gov.uk', target: :blank, rel: 'noopener noreferrer'} has more information.
+      %p The #{link_to 'technical documentation', 'https://docs.registers.service.gov.uk'} has more information.
 
       %p Alternatively, you can download a copy of the register data directly from each register’s homepage.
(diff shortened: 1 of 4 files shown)


app/views/pages/using_registers.html.haml
```diff
@@ -39,6 +39,6 @@
         %li download either the whole or part of a register
         %li find a single record, entry or item
 
-      %p The #{link_to 'technical documentation', 'https://registers-docs-testing.cloudapps.digital', target: :blank, rel: 'noopener noreferrer'} has more information.
+      %p The #{link_to 'technical documentation', 'https://docs.registers.service.gov.uk', target: :blank, rel: 'noopener noreferrer'} has more information.
```
gidsg: why `noreferrer` surely this is useful information to know how people are getting to the docs?

danmitchell-: We don't need this :)

app/views/registers/show.html.haml
```diff
@@ -20,7 +20,7 @@
       .column-two-thirds
         .panel.panel-border-wide
           %span.phase-banner In progress
-          %p This data is in progress and it’s not ready for use. #{link_to 'Contact the team', 'https://registers.service.gov.uk/support', target: :blank} to give us feedback.
+          %p This data is in progress and it’s not ready for use. #{link_to 'Contact the team', 'https://registers.service.gov.uk/support'} to give us feedback.
```
gidsg: can we use a route for this rather than absolute URL?

PR #277, opened 2018-05-01
Revert tabs

danmitchell-:
Remove tabs from beta-assessment-v2 branch. Half way house between live and what we want to have with accordions.

### Guidance to review
i.e. `/registers/country`


gidsg:
Can we make this just the revert and do the other content tweaks separately?

PR #278, opened 2018-05-01
Content updates

danmitchell-:
Content updates


app/views/pages/services_using_registers.html.haml
```diff
@@ -17,82 +15,82 @@
...
             %th E-petitions
             %td Parliamentary Digital Service
-            %td= link_to 'Country', 'https://country.register.gov.uk', target: :blank
+            %td= link_to 'Country', '/registers/country'
           %tr
             %th GOV.UK Pay
             %td Cabinet Office
             %td
-              = link_to 'Country', 'https://country.register.gov.uk', target: :blank
+              = link_to 'Country', '/registers/country'
```
gidsg: can we resolve these using the router paths?

danmitchell-: Maybe but im not sure its worth it...

PR #286, opened 2018-05-02
Move zendesk config into rails config

gidsg:
### Context
Trying to get all config in one place

Move zendesk config into rails config

### Guidance to review
This is a refactor, should work as before.


danmitchell-:
We really should be using secrets for these keys

PR #289, opened 2018-05-04
Update ga

danmitchell-:
### Context
Improve Google Analytics tracking on new user journeys

- Update gaap-analytics script
- Add tracking to footer links
- Add tracking to homepage inline link
- Add tracking to homepage hero CTA

### Guidance to review
@dan-gilbert


app/assets/javascripts/gaap-analytics.js
```diff
@@ -1,6 +1,6 @@
 // Taken from https://raw.githubusercontent.com/alphagov/gaap-analytics/master/build/gaap-analytics.js
```
gidsg: is this just the latest version of an upstream dependency?

danmitchell-: Yeah. This is the latest version which includes a few dependencies. One day we'll setup npm for the app so we can pull this package from npmjs.com but for now its a copy and paste job.

app/views/pages/home.html.haml
```diff
@@ -53,4 +53,11 @@
       .grid-row
         .column-full
           %h2.heading-medium.content-section__title Get started
-          %p Search all the #{link_to 'registers you can use', registers_path} to see what government data is available.
+          %p{data: {"click-events" => "", "click-category" => "Content", "click-action" => "Navigation Link Clicked"}} Search all the #{link_to 'registers you can use', registers_path} to see what government data is available.
+
+= content_for :javascript do
```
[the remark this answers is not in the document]
dan-gilbert: As there are multiple ways to get to the /registers page from the homepage, we're sending a virtual page view here to be able to identify sessions where the button is used.  We'll then use this virtual page path in a goal funnel in Google Analytics.

PR #298, opened 2018-05-10
Add tracking to table sorting and load more

danmitchell-:
### Context
Tracking on register table sorting and load more

Add event tracking to table sorting links

### DIFF
--- a/app/views/kaminari/_paginator.html.haml
+++ b/app/views/kaminari/_paginator.html.haml
@@ -1,5 +1,5 @@
 = paginator.render do
-  %ul.pager-items
+  %ul.pager-items{"data-click-events" => "", "data-click-category" => "Register Table", "data-click-action" => "Pagination"}
     = prev_page_tag unless current_page.first?
     - each_page do |page|
       - if page.display_tag?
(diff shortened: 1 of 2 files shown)


app/views/kaminari/_paginator.html.haml
```diff
@@ -1,5 +1,5 @@
 = paginator.render do
-  %ul.pager-items
+  %ul.pager-items{"data-click-events" => "", "data-click-category" => "Register Table", "data-click-action" => "Pagination"}
```
gidsg: does this empty string `"data-click-events" => ""` do anything, can we just remove it?

danmitchell-: Initializes the JS 👍

gidsg: if you use `nil` instead of `""` does it generate the same markup?

danmitchell-: Doesn't render in the DOM if you use nil.

gidsg: how about `"data-click-events" => true`?

PR #301, opened 2018-05-11
use rocket instead of colon

gidsg:
### Context
see https://github.com/openregister/registers-frontend/pull/300

Previous attempt set an invalid header as `:` meant header was interpreted as a symbol rather than a string. 🤦‍♂️ 

### Guidance to review
Should set valid cache control header on production


danmitchell-:
🤞

PR #306, opened 2018-05-16
Add download request

danmitchell-:
### Context
Collect user details when downloading a register

Add form in front of the download links

### Guidance to review
Visit any register page, open the "access this data" accordion and then click download

# After
screen shot 2018-05-16 at 09 28 17


app/controllers/download_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+    end
+  end
+
+  def success; end
+
+private
+
+  def download_user_params
+    params.require(:download_user).permit(
+      :email_gov,
```
danmitchell-: this should be `emai`

app/models/concerns/form_validations.rb
```diff
@@ -0,0 +1,13 @@
+require 'active_support/concern'
+
+module FormValidations extend ActiveSupport::Concern
```
danmitchell-: Can `extend ActiveSupport::Concern` be on a newline to fix the indentation of this file?

michaelabenyohai:
The build is failing @gidsg @danmitchell-

danmitchell-:
The schema.rb is missing from your commit @gidsg

michaelabenyohai:
Should both download links force a download of a file rather than just ink to a URL?

michaelabenyohai:
If a user does multiple downloads using the same details we put those same details in the database multiple times. Is this a conscious decision? 

I see that this would be useful if we were going to use this database table to identify how often a particular user downloads and for which registers, but to do that we would also need the name of the register they downloaded written to the database. Unless "downloads per person per register" is available from GA? But if it's available from GA then I'm thinking that we don't need to record multiple downloads by the same user in the database, only the first time?

gidsg:
@danmitchell- I think all the schema/migration changes were removed in: https://github.com/openregister/registers-frontend/pull/306/commits/1d77b7cd9d3975d328e0cf1a81a482dcd7b6ddab

danmitchell-:
Okay cool, as long as we squash this when merge because its a very messy commit history :)

PR #310, opened 2018-05-18
Update api key flow

danmitchell-:
### Context
Users are struggling with the "next step" once they have created an API key.

Add curl example to API key show page

### Guidance to review
Create a new API key from a register page

# After - coming from a register page
screen shot 2018-05-18 at 10 14 34

# After - going straight to `/api_users/new`
screen shot 2018-05-18 at 12 50 27

### DIFF
--- a/app/views/api_users/show.html.haml
+++ b/app/views/api_users/show.html.haml
@@ -15,7 +15,7 @@
         %p
           %code.code.code-curl
             curl #{@register.url}/records.json?page-size=5000 --header "Authorization: #{@api_key}"
-        %p This data is also avaliable in other formats. Read more in our #{link_to 'technical documentation', 'https://docs.registers.service.gov.uk/#choose-a-response-format'}.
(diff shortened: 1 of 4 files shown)


app/views/api_users/show.html.haml
```diff
@@ -8,18 +8,14 @@
       %code.code.code-large
         = @api_key
 
-      %p.font-xsmall Your API key has also been emailed to you. Your key can be used for all registers.
+      %p Your API key has also been emailed to you. Your key can be used for all registers.
+
+      %h3.heading-medium Get this register as JSON
+      %p
+        %code.code.code-curl
+          curl https://#{session[:register].present? ? session[:register] : 'register-name'}.register.gov.uk/records.json?page-size=5000 --header "Authorization: #{@api_key}"
```
gidsg: This will not work if they register from an alpha register as the URL will be wrong.

gidsg: the URL is stored in the model as `register.url`

danmitchell-: Thanks Gids, i'm refactoring this now 👍

PR #323, opened 2018-05-25
Add feedback form

danmitchell-:
### Context
We want to make giving feedback per register easier

Add feedback form per register

### Guidance to review
Visit any register page and complete the feedback form

# After
screen shot 2018-05-25 at 17 00 05


app/services/zendesk_feedback.rb
```diff
@@ -18,7 +18,7 @@ def send_feedback(params)
       requester: {
         # Passing Anonymous down to ZenDesk as name is not always required in our forms
         name: params[:name].present? ? params[:name] : 'Anonymous',
-        email: params[:email]
+        email: params[:email].present? ? params[:email] : 'registerteam+anonymous@digital.cabinet-office.gov.uk'
```
gidsg: i take it this is mandatory in zendesk?

danmitchell-: Unfortunately, yes

michaelabenyohai:
Only comment is that after submitting the URL is `/registers/country#subsection_content_3` - should we make this something more meaningful?

dankmitchell:
Yeah, good shout @michaelabenyohai!

PR #326, opened 2018-05-30
validate format of feedback email, use shared validator

gidsg:
### Context
previously we did not validate format of feedback email address (when populated)

validate feedback email, use shared custom validator

### Guidance to review
invalid feedback email should report error

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,7 +7,7 @@ end
 
 ruby File.read(".ruby-version").chomp
 
-gem 'email_validator', '~> 1.6'
+gem 'email_validator', github: 'mailtop/email_validator'
 gem 'rails', '~> 5.1.5'
 gem 'pg', '~> 0.18'
 gem 'puma', '~> 3.0'
--- a/Gemfile
+++ b/Gemfile
@@ -7,6 +7,7 @@ end
 
 ruby File.read(".ruby-version").chomp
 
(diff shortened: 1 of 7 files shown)


edent:
That's not the right way to validate an email address.

Fails on:
* Unicode TLDs like .中国
* Unicode domains like http://☮.com
* Usernames with apostrophes like `baba.o'riley@example.com`

Succeeds on:
* Domains with invalid characters like `test@exampl_.com`

gidsg:
thanks for your feedback @edent do you know of a regex which matches these cases? Or would you suggest it's not something that can be done in a regex?

gidsg:
@edent ...I have updated it now to use the `email_validator` gem which should support valid email addresses as described in: https://github.com/balexand/email_validator/blob/master/spec/email_validator_spec.rb

edent:
That gem looks long abandoned.  The only true way to validate an email address is to send an email to it.

dankmitchell:
@gidsg maybe we should either use this fork (https://github.com/mailtop/email_validator) or not bother at all.

Verify-frontend is using this gem and I've used it successfully in the past.

gidsg:
@edent re: https://github.com/openregister/registers-frontend/pull/326#issuecomment-393136359 I understand your concern about sending a validation email to the user, however, this is not within the scope of this particular PR. We may consider this as part of future feature development.

PR #329, opened 2018-05-30
Add email hint text and remove validation for other fields

dankmitchell:
### Context
We want to remove as much friction as possible when users want to use our API

Remove mandatory fields other than email

### Guidance to review
Visit `/create-api-key` and complete the form


gidsg:
this value needs to be sent in `post_to_endpoint`

dankmitchell:
It is already @gids, unless I’m missing something

app/controllers/concerns/form_helpers.rb
```diff
@@ -16,23 +16,13 @@ def government_orgs_local_authorities
...
 
 def post_to_endpoint(user, endpoint = 'users')
   @user = { email: user.email,
-            department: user.department,
-            non_gov_use_category: user.non_gov_use_category,
-            is_government: user.is_government,
-            register: user.try(:register) }
+            is_government: user.is_government_boolean,
+            register: user.try(:register),
+            contactable: user.is_contactable_boolean }
```
gidsg: this needs to be a `try` unless it is also used by the download form.

app/views/api_users/_form.html.haml
```diff
@@ -3,24 +3,16 @@
     - fieldset.radio_input :yes, panel_id: 'is_government_panel'
     - fieldset.radio_input :no, panel_id: 'not_government_panel'
   - f.revealing_panel :is_government_panel do |panel|
-    = f.collection_select :department, government_orgs_local_authorities, :first, :last, { include_blank: true }
     = f.text_field :email_gov, type: 'email'
+    = f.radio_button_fieldset :contactable, inline: true
```
gidsg: if the user does not select an option for this we need to show a validation error. It is not optional.

PR #336, opened 2018-06-04
Update images on homepage

dankmitchell:
Update homepage images

### Guidance to review
# After
screen shot 2018-06-04 at 12 59 10


gidsg:
Is there a reason the images are these dimensions? As they seem to always be being scaled down in the browser.

dankmitchell:
Supplied by @conordelahunty...

gidsg:
I think these should be scaled down to 465x310

dankmitchell:
Done @gidsg. I really think we need SVG's at some stage

PR #339, opened 2018-06-05
Add position sorting to registers

dankmitchell:
### Context
We want to control the ordering off registers on the `/get-data` page

Allow admins to drag and drag the list of registers

### Guidance to review
# After
drag-and-drop


app/models/register.rb
```diff
@@ -1,4 +1,7 @@
 class Register < ApplicationRecord
+  include PgSearch
+  pg_search_scope :search_for, against: { name: 'A' }
```
gidsg: what is this doing, what is the magic value `'A'`?

app/views/admin/registers/_register.html.haml
```diff
@@ -1,6 +1,4 @@
-%tr
-  %td
-    = register.name
-  %td= register.register_phase
-  %td
-    = link_to 'Edit', edit_admin_register_path(register)
+%li{id: dom_id(register)}
```
gidsg: can we add some explanation e.g. Drag to re-order registers by popularity? Can we add some icon to make it clearer these are draggable?

app/controllers/registers_controller.rb
```diff
@@ -2,8 +2,18 @@
 
 class RegistersController < ApplicationController
   def index
-    @search = Register.available.ransack(params[:q])
-    @registers = @search.result.where(register_phase: 'Beta').sort_by_phase_name_asc.by_name
+    if params[:q].present?
+      @registers = Register.available.in_beta.search_for(params[:q])
+    elsif params[:sort].present?
```
gidsg: I think this logic can be moved into the model with an optional parameter. i.e. the pattern in: https://github.com/openregister/registers-frontend/blob/2bff975b668152d0858fc1e1a72a1860c86bcffb/app/models/concerns/search_scope.rb#L4

app/models/register.rb
```diff
@@ -10,9 +13,11 @@ class Register < ApplicationRecord
   validates :slug, uniqueness: true
 
   scope :by_name, -> { order name: :asc }
+  scope :by_position, -> { order position: :asc }
```
gidsg: can we call this `by_popularity` as it closer matches the UI and intention?

PR #343, opened 2018-06-08
Honeypot field on feedback

dankmitchell:
### Context
Spam coming through to ZenDesk is causing unnecessary overhead to manage

Add [invisible_captcha](https://github.com/markets/invisible_captcha) i.e. honeypot field to feedback form to fight spam

### Guidance to review

 If you inspect the element, untick the display-none which will show the field, out something the text_field and submit the form. The response should be `no content` but this [customisable](https://github.com/markets/invisible_captcha#usage) i.e. we could redirect to the homepage.

#### The DOM output
```
<div class="spam_1528453644">
<style media="screen">
.spam_1528453644 {position:absolute!important;height:1px;width:1px;overflow:hidden;}
</style>
<label for="feedback_spam">If you are a human, ignore this field</label>
<input type="text" name="feedback[spam]" id="feedback_spam" tabindex="-1">
</div>
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -65,6 +65,8 @@ gem 'httparty', '~> 0.16.2'
 # Canonical meta tag
 gem 'canonical-rails'
 
+gem 'invisible_captcha'
+
 group :development, :test do
   gem 'govuk-lint', '~> 3.8'
   gem 'pry-byebug'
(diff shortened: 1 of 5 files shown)


accessiblewebuk:
If this hasn't already been tested for accessibility, would recommend that it is, in particular for screen reader software (preferably JAWS with IE, NVDA with Firefox and VoiceOver on iOS which are the most commonly used combinations - see https://accessibility.blog.gov.uk/2016/11/01/results-of-the-2016-gov-uk-assistive-technology-survey/)

PR #344, opened 2018-06-08
Add single record resolver and record linking 🔗

gidsg:
### Context
Previously the frontend did not show any links between records.

* Adds single record key lookup using `record` query parameter
* Adds links for register type fields and CURIEs
* resolves all links for cardinality N fields
single record linking

### Guidance to review
All links should resolve as long as corresponding registers are loaded

### DIFF
--- a/app/controllers/registers_controller.rb
+++ b/app/controllers/registers_controller.rb
@@ -30,7 +30,7 @@ class RegistersController < ApplicationController
 
   def field_link_resolver(field, field_value, register_slug = @register.slug)
     single_resolver = lambda { |f, fv|
-      if f['register'].present? && f['register'] != register_slug
+      if f['register'].present? && f['field'] != register_slug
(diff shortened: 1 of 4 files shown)


app/controllers/registers_controller.rb
```diff
@@ -25,6 +28,25 @@ def show
     @feedback = Feedback.new
   end
 
+  def field_link_resolver(field, field_value, register_slug = @register.slug)
+    single_resolver = lambda { |f, fv|
+      if f['register'].present? && f['register'] != register_slug
```
michaelabenyohai: This will prevent linking in any register that links to itself. See `allergen` register, `allergen-group` field - this field is a link but not resolving.

Perhaps the logic should check whether the field has the same name as the register rather than whether the `register` property has the same name.

gidsg:
closing awaiting design input from @conordelahunty

PR #345, opened 2018-06-11
Remove backlog registers

dankmitchell:
- remove backlog registers from `/registers-in-progress` page
- remove `/suggest-register` flow
- remove `/request-register` flow
- remove legacy request and suggest register code
- Add 301 redirect for the above to redirect to `/support`

### Guidance to review
- check `/registers-in-progress`
- check `/suggest-register`
- check `/request-register`

🔥


config/routes.rb
```diff
@@ -19,19 +18,14 @@
...
-
   get '/registers-in-progress', to: 'registers#in_progress'
 
-  resources :suggest_registers, path: 'suggest-register', except: %i[show edit]
-  get 'suggest-register/complete', to: 'suggest_registers#complete'
+  resources :api_users, path: 'create-api-key'
 
-  resources :request_registers, path: 'request-register', except: %i[show edit]
-  get 'request-register/complete', to: 'request_registers#complete'
+  get '404', to: 'errors#not_found'
```
gidsg: can we move `404` and `500` to the bottom with the redirects as I think they make more sense there.

app/views/registers/in_progress.html.haml
```diff
@@ -42,37 +42,11 @@
               %th{role: "columnheader", scope: "col"}
                 Status
           %tbody
-            - @upcoming_registers.each_with_index do |register, index|
+            - @registers.each_with_index do |register, index|
               - index = index + 1
```
gidsg: seems like this `index` is only used for a `data-title` attribute in the template, can we just delete it or does it have a purpose?

dankmitchell: Removed 👍

config/routes.rb
dankmitchell: Compromised and put them above `/admin` routes.

PR #347, opened 2018-06-12
Add cookies page inline with our service

dankmitchell:
### Context
Show our users what cookies we use, what they are for and how long until they expire

Add cookie policy

### Guidance to review
`/cookies`


gidsg:
localhost_3000_cookies

PR #348, opened 2018-06-13
Only show registers with records in alpha and beta in the sitemap

dankmitchell:
### Context
The sitemap was referencing registers that aren't good enough to display on Google or to our users

Only show registers in `alpha` and `beta` that also have `records` in the sitemap

### Guidance to review
`/sitemap.xml`


app/controllers/sitemaps_controller.rb
```diff
@@ -1,5 +1,5 @@
 class SitemapsController < ApplicationController
   def show
-    @registers = Register.sort_by_phase_name_asc.by_name
+    @registers = Register.where(register_phase: %w[Alpha Beta]).has_records.sort_by_phase_name_asc.by_name
```
gidsg: no need to sort these.

PR #352, opened 2018-06-14
Add meta description and seo title for register pages

dankmitchell:
### Context
We want to add a Google friendly seo title and meta description

Add additional fields into the admin to enable the ability of an seo title and meta description

### Guidance to review
`/admin` edit any register to add an seo title and meta description

### DIFF
--- a/app/views/registers/show.html.haml
+++ b/app/views/registers/show.html.haml
@@ -1,5 +1,5 @@
-- @page_title = "#{@register.seo_title.present? ? @register.seo_title : @register.register_name + ' register'}"
-- @page_description = "#{@register.meta_description.present? ? @register.meta_description : @register.register_description}"
+- @page_title = "#{[@register.seo_title, @register.register_name].find(&:present?)}"
+- @page_description = "#{[@register.meta_description, @register.register_description].find(&:present?)}"
 - content_for :body_classes, 'register-show'
 
 %nav.breadcrumbs{role: "Navigation", "aria-label" => "Breadcrumb"}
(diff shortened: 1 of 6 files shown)


app/views/registers/show.html.haml
```diff
@@ -1,5 +1,5 @@
-- @page_title = "#{@register.register_name} register"
-- @page_description = @register.register_description
+- @page_title = "#{@register.seo_title.present? ? @register.seo_title : @register.register_name + ' register'}"
```
gidsg: I think you could express this more concisely as `[@register.seo_title, @register.register_name].find(&:present?)`

app/views/registers/show.html.haml
```diff
@@ -1,5 +1,5 @@
...
+- @page_description = "#{@register.meta_description.present? ? @register.meta_description : @register.register_description}"
```
gidsg: I think we should also consider using this for the secondary heading span within the H1 as I don't think meta description alone will have a big SEO effect.

dankmitchell: The meta description is for google search only.

app/views/registers/show.html.haml
```diff
@@ -1,5 +1,5 @@
-- @page_title = "#{@register.seo_title.present? ? @register.seo_title : @register.register_name + ' register'}"
-- @page_description = "#{@register.meta_description.present? ? @register.meta_description : @register.register_description}"
+- @page_title = "#{[@register.seo_title, @register.register_name].find(&:present?)}"
```
gidsg: do you still want to append `register` on the end? Or only for the `register_name` maybe?

dankmitchell: Yes, you are right.

PR #354, opened 2018-06-15
Remove scheduler and queue apps from research manifest

michaelabenyohai:
### Context
We don't run these for the research app as we can manually load registers when we need them.

Remove scheduler and queue apps from research manifest.

### DIFF
--- a/manifest-research.yml
+++ b/manifest-research.yml
@@ -12,13 +12,3 @@ applications:
 - name: registers-frontend-research
   <<: *defaults
   instances: 2
-- name: registers-frontend-queue
-  <<: *defaults
-  instances: 1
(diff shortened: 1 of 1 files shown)


dankmitchell:
@michaelabenyohai they are referenced in the deploy script https://github.com/openregister/registers-frontend/blob/539e4521ee96b001fcdbc40de888e9ad8699e101/bin/deploy#L5

michaelabenyohai:
@dankmitchell I hadn't realised. Do we use this script when deploying to the research environment though? I thought it was only used by Travis?

dankmitchell:
That's true actually. Manual deploys will be fine

PR #364, opened 2018-06-25
Add route and view for fields

dankmitchell:
### Context
We want to show users the field descriptions

- Render field description in own view for non-js users
- Render field description in dialog for js users

### Guidance to review
Click `?` icon next to any field name in table

Accessibility critea - https://gist.github.com/hannalaakso/2641fc16d2158e60d551cd9da960b5da

# After -JS Users
screen shot 2018-06-25 at 18 13 57

# After - NON-JS Users
screen shot 2018-06-25 at 18 14 05

### DIFF
--- a/app/views/fields/_field.html.haml
+++ b/app/views/fields/_field.html.haml
@@ -1,2 +1,2 @@
-%h1.heading-large= @field.data['field'].tr('-', ' ').humanize
-%p= @field.data['text']
+%p.modal-dialog__inner__text.bold-small= @field.data['field'].tr('-', ' ').humanize
+%p.modal-dialog__inner__text= @field.data['text']
--- a/app/views/fields/_field.html.haml
+++ b/app/views/fields/_field.html.haml
@@ -1,0 +1,2 @@
+%h1.heading-large= @field.data['field'].tr('-', ' ').humanize
+%p= @field.data['text']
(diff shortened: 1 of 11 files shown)


gidsg:
question mark is overlapping link in this case
screen shot 2018-06-26 at 11 36 56

gidsg:
Can we make it also close the dialog box if you click anywhere in the greyed out area?

app/views/fields/_field.html.haml
```diff
@@ -0,0 +1,2 @@
+%h1.heading-large= @field.data['field'].tr('-', ' ').humanize
```
gidsg: I don't think this should be an H1 within the overlay as there would then be multiple H1s on the page.

app/views/registers/show.html.haml
```diff
@@ -86,7 +86,9 @@
                     %thead
                       %tr{data: {"click-events" => true, "click-category" => "Register Table", "click-action" => "Sort"}}
                         - @register.register_fields.each do |field|
-                          %th{class: "#{field['field']}"}= field['cardinality'] == '1' ? records_table_sort_link(field['field'], request.query_parameters) : field['field']
+                          %th{class: "#{field['field']}"}
+                            = field['cardinality'] == '1' ? records_table_sort_link(field['field'], request.query_parameters) : field['field']
+                            = link_to '?', register_field_url(@register.slug, field['field']), class: 'info-icon', remote: true, "aria-label" => "What does #{field['field']} mean?"
```
gidsg: Not sure of the merits of `aria-label` vs hidden off-screen text?

dankmitchell: Replaced with a visually hidden span element 👍

app/views/fields/_field.html.haml
dankmitchell: Awaiting feedback from Connor before changing this but yes, I agree, it shouldn't be a H1 in the dialog.

PR #372, opened 2018-06-28
Update registers index

dankmitchell:
### Context
Current solution on the index view isn't scalable and doesn't work for long register name or descriptions.

Re-Introduce the table and add a `featured` checkbox so we can control a subset of "featured" registers above the full table of registers

### Guidance to review
`/registers`

# After
screen shot 2018-06-28 at 13 44 23
screen shot 2018-06-28 at 13 44 44


gidsg:
If we are not using the popularity stuff anymore, can that be deleted?

dankmitchell:
The order of the `featured` registers are still based on popularity @gidsg

app/views/registers/index.html.haml
```diff
@@ -29,29 +29,42 @@
...
-              .column-one-third
-                = link_to register_path(register.slug), class: 'register-block' do
-                  %h3.register-block__name= register.register_name
-                  %p.register-block__description= register.register_description
-                  - if register.register_authority.present?
-                    %div{class: "govuk-organisation-logo #{register.register_authority.data['name'].parameterize}"}
-                      %div{class: "logo-container #{crest_class_name(register.register_authority.data['name'].parameterize)}"}
-                        %span= register.register_authority.data['name']
+        .grid-row
+          .column-full
```
gidsg: Can we make this block conditional on `@registers.featured.any?` ?

dankmitchell: Done

PR #374, opened 2018-06-28
Only show user_records in table and for pagination

dankmitchell:
### Context
ISA Register will go into `ready to use` before the data is available 

Render table with only fields i.e. table headers

### Guidance to review
Create a register with only fields i.e. system entries

# After
screen shot 2018-06-27 at 14 31 02

### DIFF
--- a/app/controllers/registers_controller.rb
+++ b/app/controllers/registers_controller.rb
@@ -23,6 +23,7 @@ class RegistersController < ApplicationController
   def show
     @register = Register.find_by_slug!(params[:id])
     @records = recover_records(@register.fields_array, params)
+    @is_empty_register = !@register.records.where(entry_type: 'user').any?
     @feedback = Feedback.new
   end
 
@@ -57,6 +58,7 @@ private
     sort_direction = params[:sort_direction] ||= 'asc'
 
     @register.records
+             .where(entry_type: 'user')
              .search_for(fields, params[:q])
              .status(params[:status])
              .sort_by_field(sort_by, sort_direction)
--- a/app/controllers/registers_controller.rb
+++ b/app/controllers/registers_controller.rb
@@ -23,7 +23,7 @@ class RegistersController < ApplicationController
   def show
     @register = Register.find_by_slug!(params[:id])
     @records = recover_records(@register.fields_array, params)
-    @is_empty_register = !@register.records.where(entry_type: 'user').any?
+    @is_empty_register = @register.records.where(entry_type: 'user').none?
     @feedback = Feedback.new
   end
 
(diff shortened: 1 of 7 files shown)


app/controllers/registers_controller.rb
```diff
@@ -57,6 +58,7 @@ def recover_records(fields, params)
     sort_direction = params[:sort_direction] ||= 'asc'
 
     @register.records
+             .where(entry_type: 'user')
```
dankmitchell: Could keep the scope and replace this with `@register.records.user_records`

app/controllers/registers_controller.rb
```diff
@@ -23,6 +23,7 @@ def in_progress
   def show
     @register = Register.find_by_slug!(params[:id])
     @records = recover_records(@register.fields_array, params)
+    @is_empty_register = !@register.records.where(entry_type: 'user').any?
```
dankmitchell: Could keep the scope and replace this with `@is_empty_register = !@register.records.user_records.any?`

or maybe `@is_empty_register = @register.records.user_records.nil?`

app/controllers/registers_controller.rb
michaelabenyohai: We weren't using the system_records scope anywhere so thought it seemed unnecessary to have user_records and system_records scoped? Now records is the same as what the user_records scope was.

app/controllers/registers_controller.rb
michaelabenyohai: The problem with this is that the `user_records` scope was influenced by query parameters and filtering. The definition of whether a register is empty is different to whether `@register.records.user_records.nil?` is true. This could be true but it's not an empty register. It just happens that someone is filtering at the time. It seemed less confusing to separate the two variables - `@records` is influenced by filtering/searching but `@is_empty_register` is not. `@is_empty_register` can be false when the `@records` is empty.

michaelabenyohai: There was also a bug where the search bar disappears when you searched for something that resulted in no records. e.g. search for "bla" in the country register, the search box disappears.  This isn't possible to fix if `@is_empty_register` depends on `@records`.

app/controllers/registers_controller.rb
```diff
@@ -23,6 +23,7 @@ def in_progress
   def show
     @register = Register.find_by_slug!(params[:id])
     @records = recover_records(@register.fields_array, params)
+    @is_empty_register = @register.records.where(entry_type: 'user').none?
```
gidsg: we can move this into the model `/app/models/register.rb`

e.g.
```
  def empty_register?
    records.where(entry_type: 'user').none? 
  end
```
then in the template `@register.empty_register?`

PR #377, opened 2018-06-29
use ilike query for register search

gidsg:
### Context
previously we only matched on full matches of words in register names

allow fuzzy match search

### Guidance to review
fuzzy search e.g. `gov` should find results

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -30,9 +30,6 @@ gem 'govuk-registers-api-client', '~> 1.2.2'
 
 gem 'autoprefixer-rails'
 
-# Search
-gem 'pg_search'
-
 # Zendesk
 gem 'zendesk_api'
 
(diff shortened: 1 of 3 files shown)


arnau:
Is this a quickwin for case insensitive partial matches? Postgres have more advanced fulltext capabilities that I thought were half-way in via `pg_search`.

dankmitchell:
We can do a lot with postgres and `pg_search`

https://github.com/Casecommons/pg_search#trigram-trigram-search

Need the `pg_trgm` extension adding to the schema.rb and `:using => :trigram` added to the `pg_search_scope` method

gidsg:
> Is this a quickwin for case insensitive partial matches?

Yes. That is the intention and also that we are not using any of the other features of `pg_search` I don't think it's worth including currently.

PR #379, opened 2018-07-05
add register search view

gidsg:
### Context
Previously we only searched on register `name`
https://trello.com/c/l7SSYCv3/1976-pipeline-search-return-results-based-on-matches-on-the-description-text

Search across 
* register name (humanized ID)
* `register-name` (friendly name)
* register description

### Guidance to review
As this is quite a complex SQL query this PR creates a Postgres View and uses [scenic](https://github.com/thoughtbot/scenic) to make it available to Rails model. Let me know if you feel strongly against this approach and would rather denormalise or use plain subqueries in code rather than a view. 
*H/T @nacnudus helping me with the SQL query* 🔥

### DIFF
--- a/app/models/register_search_result.rb
+++ b/app/models/register_search_result.rb
@@ -1,0 +1,12 @@
+class RegisterSearchResult < ActiveRecord::Base
+  self.primary_key = 'register_id'
+  scope :search, ->(search_term) { where(*['name ILIKE ? OR register_name ILIKE ? OR register_description ILIKE ?'].fill("%#{sanitize_sql_like(search_term)}%", 1..3)) }
+
+  def readonly?
+    true
+  end
+
+  def self.search_for(search_term)
+    Register.where(id: search(search_term).ids)
+  end
+end
--- a/app/models/register_search_result.rb
+++ b/app/models/register_search_result.rb
@@ -1,9 +1,13 @@
-class RegisterSearchResult < ActiveRecord::Base
+class RegisterSearchResult < ApplicationRecord
   self.primary_key = 'register_id'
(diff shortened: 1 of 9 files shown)


app/models/register_search_result.rb
```diff
@@ -0,0 +1,12 @@
+class RegisterSearchResult < ActiveRecord::Base
+  self.primary_key = 'register_id'
+  scope :search, ->(search_term) { where(*['name ILIKE ? OR register_name ILIKE ? OR register_description ILIKE ?'].fill("%#{sanitize_sql_like(search_term)}%", 1..3)) }
```
arnau: This should be approached with postgres full-text capabilities for performance sake. Either in this PR or in another one as long as the delay is not too much.

db/schema.rb
```diff
@@ -90,4 +90,20 @@
     t.datetime "password_reset_sent_at"
   end
 
+
+  create_view "register_search_results",  sql_definition: <<-SQL
```
arnau: Is this a materialised view or not? In other words, is there any delay on reflecting changes when the underlying data changes?

arnau: I answered this question reading scenic readme 👍

app/models/register_search_result.rb
arnau: I thought pg_search wouldn't see the JSONB given you are using a view. Ignore me then.

gidsg: good point. I have not tried using `pg_search` to query the view, but I'm not sure it buys us a lot even if it is supported. I will take a look.

gidsg: @arnau after some testing I have found for a table of this small size Postgres will elect to do a scan even when an index is available, so there is no benefit in adding one at this point. We could add a materialised view and re-index it if we think the extra overhead of refreshing it is worth the performance benefit.

arnau: Ok thanks for looking into it 👍

PR #386, opened 2018-07-12
Require records to exist before linking to a register

MatMoore:
### Context
When new registers are being developed, the frontend can get into a state where there are registers than reference records in registers that do not show up on the frontend. So the user follows a link to a 404.

This happens because:

- Populating the frontend is an asynchronous process
- Registers have to be whitelisted to show up in the frontend

Trello: https://trello.com/c/FUnf29SI/2621-link-resolver-should-only-link-to-available-registers

The change is to only link to a record or register if the register being linked to has records in it. If it doesn't have records, just show the value without creating a link.

### Guidance to review
I've tried to factor out the link resolution code from the controller by creating a new class, so that I can extract some of the business logic to helper methods. I'm not sure if this is the best way of implementing it, so I've kept this as a separate commit - let me know if you'd prefer to revert it.

I also considered making it a rails helper, but it meant I couldn't extract methods out without having to keep passing around the 4 state variables, so I abandoned that approach.

### DIFF
--- a/app/lib/link_resolver.rb
+++ b/app/lib/link_resolver.rb
@@ -12,9 +12,13 @@ class LinkResolver
     field_name = field['field']
     field_datatype = field['datatype']
 
+    foreign_key_resolvable = field_register_slug.present? && \
+      field_name != current_register_slug && \
+      register_whitelist.include?(field_register_slug)
+
     if field_datatype == 'curie' && field_value.include?(':')
       resolve_curie(field_value)
-    elsif field_register_slug.present? && field_name != current_register_slug && register_whitelist.include?(field_register_slug)
+    elsif foreign_key_resolvable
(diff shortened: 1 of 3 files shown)


app/lib/link_resolver.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def resolve(field, field_value)
+    field_register_slug = field['register']
+    field_name = field['field']
+    field_datatype = field['datatype']
+
+    if field_datatype == 'curie' && field_value.include?(':')
+      resolve_curie(field_value)
+    elsif field_register_slug.present? && field_name != current_register_slug && register_whitelist.include?(field_register_slug)
```
gidsg: very minor but given there are quite a few `&&` conditions here, might be more readable if we split onto multiple lines?

PR #393, opened 2018-07-23
Add end_date field to Record

MatMoore:
### Context
Currently, the "end-date" of a record is accessed directly from the "data" json field, and it's type is an ISO 8601 string.

The "end-date" field name is conventionally used to indicate when a record stops being accurate. On the registers frontend we use that to implement a status filter.

This filter should take into account the actual value of the end-date field, and not just its presence. It needs to be able to handle partial datetimes, for example "2015" - which we would interpret as "2015-01-01T00:00:00Z" ([trello card](https://trello.com/c/r48dv7N1/2662-archived-filter-should-not-include-end-dates-that-are-in-the-future))

In preparation for this, add a field to all records with the expanded value of the end-date. A record is current if end_date is nil or a datetime that's in the past.

### Guidance to review
This PR should add the new field, change the worker to populate this for new records, and also update the existing records. I'll raise another PR for the actual change to the status filter.

I'm converting the string to a datetime using @arnau's ISO8601 gem: https://github.com/arnau/ISO8601

I've implemented the update as a migration, but it takes 7 minutes to run, so I'm not sure if that will cause any issues when deploying. I've disabled the transaction for this just in case it holds any database locks that would interfere with the normal operation of the app.


arnau:
@MatMoore could you give me a bit of context of why changing the database values is the chosen solution? I don't know enough how the frontend applies the end date filter so that's probably one of the things I would benefit from knowing to understand this PR better.

Also, there are some values in `end-date` in some registers that lack timezone eventhough they have a time component, have you checked it keeps the right value after the conversion?

MatMoore:
@arnau as discussed earlier, this is to allow us to do the filtering in activerecord using a named scope. This is the code I'll be updating in the next PR: https://github.com/openregister/registers-frontend/blob/master/app/controllers/registers_controller.rb#L62

I've pushed up another test that checks a record without the timezone component (using an example from the [service register](https://www.registers.service.gov.uk/registers/government-service?utf8=%E2%9C%93&q=&status=archived#records_wrapper))

I've also checked that ORJ conforms to the spec in only allowing UTC timestamps - this is implemented here: https://github.com/openregister/openregister-java/blob/master/src/main/java/uk/gov/register/util/DateFormatChecker.java#L38

So these are all the different ways of specifying a datetime:

```ruby
irb(main):002:0> ISO8601::DateTime.new("2018")
=> #<ISO8601::DateTime:0x00007fd7e2409f18 @original="2018", @date_time=Mon, 01 Jan 2018 00:00:00 +0000, @second=0.0>
irb(main):001:0> ISO8601::DateTime.new("2018-01")
=> #<ISO8601::DateTime:0x00007fd7e24482e0 @original="2018-01", @date_time=Mon, 01 Jan 2018 00:00:00 +0000, @second=0.0>
irb(main):003:0> ISO8601::DateTime.new("2018-01-01")
=> #<ISO8601::DateTime:0x00007fd7e0a12ee8 @original="2018-01-01T00:00:00", @date_time=Mon, 01 Jan 2018 00:00:00 +0000, @second=0.0>
irb(main):006:0> ISO8601::DateTime.new("2018-01-01T00:00:00Z")
=> #<ISO8601::DateTime:0x00007fd7e1a63c70 @original="2018-01-01T00:00:00Z", @date_time=Mon, 01 Jan 2018 00:00:00 +0000, @second=0.0>
```

PR #395, opened 2018-07-25
Underlines on the homepage stats are brought back.

injms:
screen shot 2018-07-25 at 11 17 46

### Context
The three numbers on the homepage were links, but it wasn't obvious that they were clickable without the underlines present.

 - Added underlines to the text by tweaking the application CSS
 - Returned line height to default
 - Made the text slightly larger for increased legibility
 - Avoid one word being on it's own line by using a non-breaking space

### Guidance to review
None.

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -313,8 +313,9 @@ table {
     font-size: 16px;
     text-decoration: none;
 
-    .data-item.bold-xsmall {
+    .data-item.bold-small {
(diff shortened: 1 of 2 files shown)


injms:
@MatMoore Sorry for suggesting the previous change that you authored shouldn't have underlines - I should've dug a bit deeper then to find this solution.

The underline is there all the time. There's a colour change on hover for non-visited links, but nothing else should change on hover.

PR #397, opened 2018-07-27
Add a new job to redownload a register

MatMoore:
### Context
We want to be able to redownload a register from scratch with zero downtime ([trello](https://trello.com/c/z7xw1LGy/2706-ability-to-force-full-reload-of-registers-in-registers-frontend-with-0-downtime-to-pick-up-metadata-changes)).

We need this in two situations:

- For discovery/alpha registers, we don't enforce the append-only nature of the register. So when we go back and change history, we need to fetch everything again. This one we can do already as we don't care about downtime.

- The java implementation stores system entries seperately, and these are not currently retrievable as part of a partial update. So as a workaround we need to rebuild the register in registers-frontend whenever new system entries are added. This we can't do now.

I've added a job which deletes a registers existing records, nullifies its root hash, and redownloads the register, inside a transaction.

The existing PopulateRegisterDataInDbJob job now calls the new job as a fallback if the register is invalid. This is equivalent to the existing behaviour - which would delete the data and then retry the job - but it does it inside a transaction.

I've factored out the part that calls the client library into a new class, and I've removed the `refresh_data` call, because I don't think we need it. The client library already calls this when you initialise a register client, so it was attempting to fetch data twice, which complicated the tests.

### Guidance to review
I've tested this locally by manually changing the root hash of a large register (charity) in register-frontend's database, then running the job. The website still works while the job is running, and after the job has run the root hash is correct and there is the correct number of records.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,8 @@ If you need to use the Zendesk service you need to add 3 environment variables
 
 Add any registers using the `/admin` UI.
 
+If you need to redownload an existing register from scratch, you can run the job `bundle exec rake registers_frontend:populate_db:force_full_register_download[SLUG]` where `SLUG` is the slug of the register.
+
 ### prod
 
 When running in production the `registers-frontend-scheduler` app periodically calls `rake registers_frontend:populate_db:fetch_later` which adds a job to a queue maintained by the `registers-frontend-queue` app. When the job runs it refreshes the data for all registers listed in the database.  
--- a/app/jobs/populate_register_data_in_db_job.rb
+++ b/app/jobs/populate_register_data_in_db_job.rb
@@ -7,7 +7,7 @@ class PopulateRegisterDataInDbJob < ApplicationJob
     RegisterDownloader.download(register)
     RegisterSearchResult.refresh
   rescue InvalidRegisterError
-    logger.info "Failed to update #{register.name}, so forcing a full reload instead"
-    RedownloadRegisterJob.perform_now(register)
+    logger.info "Root hash has changed for #{register.name}, so forcing a full reload instead"
+    ForceFullRegisterDownloadJob.perform_now(register)
   end
 end
--- a/app/jobs/populate_register_data_in_db_job.rb
+++ b/app/jobs/populate_register_data_in_db_job.rb
@@ -1,36 +1,13 @@
 class PopulateRegisterDataInDbJob < ApplicationJob
   queue_as :default
 
-  module Exceptions
-    class FrontendInvalidRegisterError < StandardError; end
-  end
-
   def perform(register)
-    Delayed::Worker.logger.info("Updating #{register.name} in database")
-    begin
-    # Initialize client and download / update data
-      registers_client_manager = RegistersClient::RegisterClientManager.new({
-        api_key: Rails.configuration.try(:registers_api_key)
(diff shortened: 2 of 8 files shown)


app/jobs/populate_register_data_in_db_job.rb
```diff
@@ -1,36 +1,13 @@
...
 
-    register.fields_array = Record.where(key: "register:#{register.slug}")
-                                  .pluck("data -> 'fields' as fields")
-                                  .first
-    register.url = register_url
-    register.save
+    RegisterDownloader.download(register)
     RegisterSearchResult.refresh
+  rescue InvalidRegisterError
+    logger.info "Failed to update #{register.name}, so forcing a full reload instead"
```
gidsg: I think this log message is ambiguous as to the failure cause, whereas this will only happen if the root hash has changed. Can we be more specific e.g.
> root hash has changed, forcing full reload

app/jobs/redownload_register_job.rb
```diff
@@ -0,0 +1,32 @@
...
+      Record.where(register_id: register.id).delete_all
+      Entry.where(register_id: register.id).delete_all
+      register.root_hash = nil
+      register.save
+
+      logger.info("Updating #{register.name} in database")
+
+      begin
+        RegisterDownloader.download(register)
+      rescue InvalidRegisterError => e
```
gidsg: I don't think it makes sense for this job to rescue from this exception. This is because this exception only gets thrown when the local root hash differs from the remote one, as we have just deleted the local root hash there will not be one so this exception should not throw, please see: https://github.com/openregister/govuk-registers-api-client/blob/76227df2db8887de40ae0882cd931686947e166c/lib/register_client.rb#L131

MatMoore: Good point - I'll just remove this rescue

app/jobs/populate_register_data_in_db_job.rb
```diff
@@ -1,36 +1,13 @@
...
-    # If register data is invalid we want to delete existing entries and records to force a full reload
-      ActiveRecord::Base.transaction do
-        Record.where(register_id: register.id).delete_all
-        Entry.where(register_id: register.id).delete_all
-        register.root_hash = nil
-        register.save
-      end
-      raise Exceptions::FrontendInvalidRegisterError, "#{register.name}: #{e}"
-    end
+    logger.info("Updating #{register.name} in database")
```
gidsg: can we check this still logs correctly when performed on the queue service?

MatMoore: Have replaced this with DelayedJob.logger

README.md
```diff
@@ -55,6 +55,8 @@ If you need to use the Zendesk service you need to add 3 environment variables
 
 Add any registers using the `/admin` UI.
 
+If you need to redownload an existing register from scratch, you can run the job `bundle exec rake populate_db:force_full_register_download[SLUG]` where `SLUG` is the slug of the register.
```
gidsg: I think we are missing `registers_frontend:` off the front of the rake task?

MatMoore: Doh. Added this in

MatMoore:
@gidsg can you give this another review?

PR #402, opened 2018-07-31
Optionally include an API key when requesting data for downloads

MatMoore:
### Context
When data is downloaded through registers frontend, it requests the data from ORJ without a key. We want this to use a key on production so that we can separate this traffic out in the analytics.

Instead of calling `open` with a URL, use HTTParty, and pass an API key if it exsits.

This sets the API key from the environment in production, but ignores it on other environments.

I originally had this just read from environment whenever its present (which is how I tested it locally), but I've changed the config to be consistent with how `config.registers_api_key` is set.

### DIFF
--- a/app/controllers/download_controller.rb
+++ b/app/controllers/download_controller.rb
@@ -39,17 +39,16 @@ class DownloadController < ApplicationController
   def success; end
 
   def download_json
-    data = open("#{@register.url}/records.json?page-size=5000", &:read)
+    data = download_format("json")
     send_data data, type: "application/json; header=present", disposition: "attachment; filename=#{@register.slug}.json"
   end
 
   def download_csv
-    data = open("#{@register.url}/records.csv?page-size=5000", &:read)
+    data = download_format("csv")
     send_data data, type: "application/csv; header=present", disposition: "attachment; filename=#{@register.slug}.csv"
   end
 
 private
-
 
   def set_register
     @register = Register.find_by_slug!(params[:register_id])
@@ -64,5 +63,22 @@ private
       :is_government,
       :register
     )
+  end
+
+  def api_key
+    Rails.configuration.try(:user_download_api_key)
+  end
+
+  def download_format(format)
+    uri = "#{@register.url}/records.#{format}?page-size=5000"
+    headers = {}
+    headers['Authorization'] = api_key if api_key.present?
+
+    response = HTTParty.get(uri, headers: headers)
+    if response.code == 200
+      response.body
+    else
+      raise "#{response.status}: #{response.message}"
+    end
   end
 end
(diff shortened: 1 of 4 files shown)


app/controllers/download_controller.rb
```diff
@@ -65,4 +64,20 @@ def download_user_params
...
+
+  def download_from_uri(uri)
+    headers = {}
+    headers['Authorization'] = api_key if api_key.present?
+
+    response = HTTParty.get(uri, headers: headers)
+    if response.code == 200
+      response.body
+    else
+      raise "#{response.status}: #{response.message}"
```
gidsg: will this return a 500 to the end user? Maybe we can create our own exception class to better capture when this happens?

app/controllers/download_controller.rb
```diff
@@ -65,4 +64,20 @@ def download_user_params
...
+    end
+  end
+
+  def api_key
```
gidsg: I found it a bit confusing that is called earlier in the file than where it is defined, can we move it above `download_from_uri`?

app/controllers/download_controller.rb
```diff
@@ -39,18 +39,17 @@ def create
...
 
   def download_json
-    data = open("#{@register.url}/records.json?page-size=5000", &:read)
+    data = download_from_uri("#{@register.url}/records.json?page-size=5000")
     send_data data, type: "application/json; header=present", disposition: "attachment; filename=#{@register.slug}.json"
   end
 
   def download_csv
-    data = open("#{@register.url}/records.csv?page-size=5000", &:read)
+    data = download_from_uri("#{@register.url}/records.csv?page-size=5000")
```
gidsg: since these arguments are the same except the format, maybe we change the argument from the URL to the format, and build the URL in the shared method?
e.g. `download_with_format('csv')`?

app/controllers/download_controller.rb
MatMoore: Yes this will return a 500 page. I'm not sure what you mean by "better capture" - do you want to show something different to the user?

gidsg: No, I just mean if we create an exception class we will see a consistent exception class in the logs we can better query for.

app/controllers/download_controller.rb
```diff
@@ -39,18 +39,17 @@ def create
...
 
   def download_json
-    data = open("#{@register.url}/records.json?page-size=5000", &:read)
+    data = download_format("json")
     send_data data, type: "application/json; header=present", disposition: "attachment; filename=#{@register.slug}.json"
   end
 
   def download_csv
-    data = open("#{@register.url}/records.csv?page-size=5000", &:read)
+    data = download_format("csv")
```
gidsg: very minor but we can use single quotes here as we are not doing string interpolation.

PR #407, opened 2018-08-07
Analytics rationalisation and update

injms:
### Context
The analytics audit showed that there was a lot of tracking where it wasn't needed, and an absence of tracking where it was needed. 

 * Events no longer fire when links the header and the footer of the page are clicked
 * Events no longer fire when a register is sorted
 * The feedback form now includes the 'No' sub-options, and only fires when the form is submitted
 * Accordions only fire an event when opened, not when closed
 * Unsuccessful searches now fire an event to let us know what query wasn't found

### Guidance to review
(Apologies for the mass of linting changes that have become inextricably linked to this PR).
This has been tested in all of the supported browsers - but extra eyes are always welcome!

### DIFF
--- a/app/assets/javascripts/accordion-with-descriptions.js
+++ b/app/assets/javascripts/accordion-with-descriptions.js
@@ -1,10 +1,9 @@
 (function (Modules) {
-  "use strict";
+  /* globals $, ga */
+  'use strict';
 
-  Modules.AccordionWithDescriptions = function() {
-
-    this.start = function($element) {
-
+  Modules.AccordionWithDescriptions = function () {
+    this.start = function ($element) {
       // Indicate that js has worked
       $element.addClass('js-accordion-with-descriptions');
 
@@ -18,7 +17,7 @@
       var $openOrCloseAllButton;
       var GOVUKServiceManualTopic = serviceManualTopicPrefix();
 
-      //addOpenCloseAllButton();
+      // addOpenCloseAllButton();
       addButtonsToSubsections();
       addIconsToSubsections();
       addAriaControlsAttrForOpenCloseAllButton();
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/accordion-with-descriptions.js
```diff
@@ -18,7 +17,7 @@
       var $openOrCloseAllButton;
       var GOVUKServiceManualTopic = serviceManualTopicPrefix();
 
-      //addOpenCloseAllButton();
```
gidsg: If this is not being used can we delete it rather than commenting it out?

gidsg:
Can we use XO linter as this is the GDS way: https://gds-way.cloudapps.digital/manuals/programming-languages/nodejs/#source-formatting-and-linting

app/assets/javascripts/accordion-with-descriptions.js
```diff
@@ -210,14 +208,14 @@
         });
       }
 
-      function openStoredSections($section) {
+      function openStoredSections ($section) {
         toggleSection($section);
         toggleIcon($section);
         toggleState($section.parent().find('.subsection__button'));
         // setOpenCloseAllText();
```
gidsg: As per above can we uncomment or delete?

app/assets/javascripts/accordion-with-descriptions.js
```diff
@@ -210,14 +208,14 @@
...
       }
 
-      function setOpenCloseAllText() {
+      function setOpenCloseAllText () {
         var openSubsections = $('.subsection--is-open').length;
         // Find out if the number of is-opens == total number of sections
```
gidsg: I know this PR has not changed this, but if this is still an open question can we prefix with `TODO: ` so it is better recognised by IDEs?

app/views/layouts/_analytics.html.erb
```diff
@@ -1,17 +1,20 @@
 <!-- Google Analytics -->
 <script>
+/* global ga, window */
```
gidsg: since this is refactoring the way we instanciate the GA JS, could we pull this out into a separate PR to split it out from the event tracking changes?

app/assets/javascripts/accordion-with-descriptions.js
injms: Sure thing - I'll bundle this into another 'neatening code' type pull request separate to this one.

app/assets/javascripts/accordion-with-descriptions.js
injms: Okay - I'll put that in a separate pull request.

app/assets/javascripts/accordion-with-descriptions.js
injms: Okay - this linter has highlighted a lot of stuff that isn't mine...

app/views/layouts/_analytics.html.erb
injms: Good point - I'll cherry pick and put in a separate pull request.

injms:
This has snowballed, as the accidental reformatting has highlighted things that neatened up. I'm closing this, and I'll put in a set of pull requests that will fix all of @gidsg comments and suggestions.

PR #408, opened 2018-08-08
Analytics rationalisation and update (take two)

injms:
### Context
The analytics audit showed that there was a lot of tracking where it wasn't needed, and an absence of tracking where it was needed.

Events no longer fire when links the header and the footer of the page are clicked
Events no longer fire when a register is sorted
The feedback form now includes the 'No' sub-options, and only fires when the form is submitted
Accordions only fire an event when opened, not when closed
Unsuccessful searches now fire an event to let us know what query wasn't found

### Guidance to review
This has been tested in all of the supported browsers - but extra eyes are always welcome!

### DIFF
--- a/app/views/registers/show.html.haml
+++ b/app/views/registers/show.html.haml
@@ -110,7 +110,6 @@
                     - if @search_term.present?
                       No results found for <strong>"#{@search_term}"</strong>
                       %script
-                        var selectedInput = '';
                         var searchFilter = document.getElementById('search_records').querySelector('input[type="radio"]:checked').parentElement.querySelector('label').innerText.toLowerCase();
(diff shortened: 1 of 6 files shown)


app/views/registers/show.html.haml
```diff
@@ -109,8 +109,14 @@
                   %p
                     - if @search_term.present?
                       No results found for <strong>"#{@search_term}"</strong>
+                      %script
+                        var selectedInput = '';
```
gidsg: in this unused? I cannot find any other reference to it?

injms: It is unused - now fixed in subsequent push.

PR #409, opened 2018-08-08
Changes Google Analytics loading method; adds versioning.

injms:
### Context
The standard method of loading Google Analytics that comes out of the box is okay - but can be made better without affecting a user's experience.

 * `analytics.js` loaded with an asynchronous `script` tag, rather than script tag being injected. Unlike the out of the box method, this allows browser to look ahead and speculatively look for any assets.
 * Versioning added as custom dimension; allows for GA dashboard to segment based on tracking setup
 * Beacon method of tracking turned off because beacon isn't supported by Internet Explorer.

### Guidance to review
None.

### DIFF
--- a/app/views/layouts/_analytics.html.erb
+++ b/app/views/layouts/_analytics.html.erb
@@ -1,17 +1,20 @@
 <!-- Google Analytics -->
 <script>
+/* global ga, window */
+window.ga = window.ga || function () { (ga.q = ga.q || []).push(arguments); };
 
-(function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
-(i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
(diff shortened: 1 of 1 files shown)


app/views/layouts/_analytics.html.erb
```diff
@@ -1,17 +1,20 @@
...
+  version: {
+    dimension: 'dimension2',
+    payload: 'v1'
+  }
+};
 
 ga('create', 'UA-86101042-3', 'auto');
 ga('set', 'anonymizeIp', true);
 ga('set', 'displayFeaturesTask', null);
-ga('set', 'transport', 'beacon');
```
gidsg: is there a downside to setting this given that GA will fallback to a supported method as per: https://developers.google.com/analytics/devguides/collection/analyticsjs/sending-hits#specifying_different_transport_mechanisms does it break IE?

injms: Didn't realise Analytics was sensible enough to provide a fallback - should've checked the reports to see if IE was registering first. Thanks for the catch! Fixed now.

PR #414, opened 2018-08-22
Email backend

MatMoore:
### Context
We are going to use mailchimp for sending regular updates about the registers service. See the big checklist here: https://trello.com/c/CRqz8itj/2694-manage-mailing-lists

This PR is purely about integrating with mailchimp for the signup journey. We will update the content on the pages separately, and if there is any frontend cleanup it would be easier to do that in another PR as well.

When the user opts in to updates, add them to a mailchimp list.

- We add them in the pending state so they receive a confirmation email
- We also send the register they signed up for so we can report by register in mailchimp

As part of this I refactored the routing and passing of query parameters between these 3 endpoints:

```
                 sign_up_index GET      /sign-up-for-updates(.:format)                            sign_up#index
                               POST     /sign-up-for-updates(.:format)                            sign_up#create
             sign_up_thank_you GET      /sign-up-for-updates/thank-you(.:format)                  sign_up#thank_you
```

You can pass these as a hash to rails' URL helpers, and you can the `&.` operator to look up the slug wherever we have a register model.

### Guidance to review
- You can test this by setting `MAILCHIMP_API_KEY` environment variable. This is available in the register pass store, but I can give you the value if it's easier
- You should get an error if the email is obviously invalid
- You should get redirected to the thank you page otherwise, and receive a mailchimp email
- You can unsusbcribe and then resubscribe again
- You can subscribe and resubscribe again (the process is the same for the user)
- Make sure the GA tracking isn't broken

Our plan is to merge this into @injms's branch asap so we don't have integration problems later. If we need to deploy the access the data flow without the mailing list part, we will just comment out the buttons/routes.

Paired with @gidsg

### DIFF
--- a/app/views/sign_up/index.html.haml
+++ b/app/views/sign_up/index.html.haml
@@ -16,9 +16,7 @@
             = flash.alert
           %p!="If the problem persists please contact  #{link_to 'support', support_path}"
 
-      = form_for @sign_up_user, url: sign_up_index_path({from: params[:from], method: params[:method]}.compact), html: { class: "js--ga-submit" } do |f|
-
-
+      = form_for @sign_up_user, url: sign_up_index_path({from: @register&.slug, method: params[:method]}.compact), html: { class: "js--ga-submit" } do |f|
         .form-group
           -# = label_tag(:email, "Email address", class: "form-label-bold")
           = f.text_field(:email, nil, class: "govuk-input govuk-!-width-one-half", required: true)
--- a/app/views/sign_up/index.html.haml
+++ b/app/views/sign_up/index.html.haml
@@ -17,9 +17,8 @@
           %p!="If the problem persists please contact  #{link_to 'support', support_path}"
 
       = form_for @sign_up_user, url: sign_up_index_path({from: @register&.slug, method: params[:method]}.compact), html: { class: "js--ga-submit" } do |f|
-        .form-group
-          -# = label_tag(:email, "Email address", class: "form-label-bold")
-          = f.text_field(:email, nil, class: "govuk-input govuk-!-width-one-half", required: true)
+        .sign-up-for-email
+          = f.email_field(:email, nil, class: "govuk-input govuk-!-width-one-half", spellcheck: false, autocapitalize: "off", required: true)
         .form-group
           = button_tag("Sign up for updates", class: "button")
 
--- a/app/views/sign_up/index.html.haml
+++ b/app/views/sign_up/index.html.haml
@@ -10,10 +10,18 @@
         %li updates to registers, the latest available registers, and upcoming registers
         %li tips and guidance on how to get the best out of registers
 
-      = form_for @sign_up_user, url: thank_you_for_signing_up_path() + @params, html: { class: "js--ga-submit" } do |f|
+      - if flash[:alert].present?
+        %div{:class => 'error-summary', :role => 'alert'}
+          %h2{:class => %w(heading-medium error-summary-heading), :id => 'error-summary-heading'}
+            = flash.alert
+          %p!="If the problem persists please contact  #{link_to 'support', support_path}"
+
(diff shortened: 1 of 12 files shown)


MatMoore:
Also, in case it's useful to anyone else: I drew this diagram of all the states a mailchimp contact can be in.

The stuff we care about in registers-frontend is the parts highlighted in yellow. 
img_20180822_141922

app/views/sign_up/thank_you.html.haml
```diff
@@ -1,3 +1,4 @@
+- console
```
gidsg: we should remove this!

config/routes.rb
```diff
@@ -10,8 +10,16 @@
   get 'privacy-notice', to: 'pages#privacy_notice', as: 'privacy_notice'
   get 'cookies', to: 'pages#cookies', as: 'cookies'
 
-  get 'sign-up-for-updates', to: 'sign_up#sign_up_for_updates', as: 'sign_up_for_updates'
-  post 'thank-you-for-signing-up', to: 'sign_up#thank_you_for_signing_up', as: 'thank_you_for_signing_up'
+  # get 'sign-up-for-updates', to: 'sign_up#sign_up_for_updates', as: 'sign_up_for_updates'
```
gidsg: deleted these commented out routes.

app/views/sign_up/index.html.haml
```diff
@@ -10,10 +10,16 @@
...
+          %h2{:class => %w(heading-medium error-summary-heading), :id => 'error-summary-heading'}
+            = flash.alert
+          %p!="If the problem persists please contact  #{link_to 'support', support_path}"
+
+      = form_for @sign_up_user, url: sign_up_index_path({from: @register&.slug, method: params[:method]}.compact), html: { class: "js--ga-submit" } do |f|
         .form-group
-          = label_tag("email-address", "Email address", class: "form-label-bold")
-          = email_field_tag("email-address", nil, class: "govuk-input govuk-!-width-one-half", required: true)
+          -# = label_tag(:email, "Email address", class: "form-label-bold")
+          = f.text_field(:email, nil, class: "govuk-input govuk-!-width-one-half", required: true)
```
gidsg: did this change affect the styling? If so @injms needs to take a look at it. Also the commented out line above should be deleted.

arnau:
@MatMoore  re. diagram: Would you mind making a digital version of it with draw.io and store the master in our google drive and the svg output in a docs folder in this repository? It would be nice to have a ADR explaining the decision.

MatMoore:
@arnau yup, I'll have a look at that now

app/views/sign_up/thank_you.html.haml
MatMoore: whoops, yeah we should - feel free to edit anything on this branch before you merge @injms @gidsg

app/views/sign_up/index.html.haml
MatMoore: Ah, yeah it did. @injms are you ok to fix this? The separate labels used to be bolded but I wasn't sure how to accomplish the same thing with `f.text_field`.

We're using the `f` object so that the fields are bound to the model object, and user input gets preserved if we re-render the same page when there's an error.

MatMoore:
Have added a digital diagram and ADR here: https://github.com/openregister/registers-frontend/pull/415

app/views/sign_up/index.html.haml
injms: Couldn't find a way of adding a class to the label other than writing a new form helper - which we might have to do in the future for the new design system, but that's a problem for then. 

The effect has been achieved with a bit of CSS. Not ideal, but it will do.

config/routes.rb
injms: ✅ Done

app/views/sign_up/thank_you.html.haml
injms: ✅ Done

PR #419, opened 2018-09-03
Registers stats on upcoming page

injms:
### Context
We need to consistently promote our available registers and usage across government across the frontend - particularly for users who may not necessarily see the homepage. 

 - Adds the three statistics near the bottom of the 'Upcoming registers' page.
 - Put the layout block in a partial HAML file so it can be included, making the one hard coded number easier to update.

### DIFF
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -4,9 +4,8 @@ class PagesController < ApplicationController
   def home
     @registers = Register.available.all
 
-    @ready_registers = Register.available.where(register_phase: 'Beta')
-    @upcoming_registers = Register.available.where.not(register_phase: 'Beta')
-    @organisation_count = @ready_registers.distinct.count(:authority)
+    @registers_available = Register.available.where(register_phase: 'Beta')
+    @organisation_count = @registers_available.distinct.count(:authority)
   end
 
   def services_using_registers; end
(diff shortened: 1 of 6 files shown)


app/controllers/pages_controller.rb
```diff
@@ -4,9 +4,8 @@ class PagesController < ApplicationController
   def home
     @registers = Register.available.all
 
-    @ready_registers = Register.available.where(register_phase: 'Beta')
-    @upcoming_registers = Register.available.where.not(register_phase: 'Beta')
-    @organisation_count = @ready_registers.distinct.count(:authority)
+    @registers_available = Register.available.where(register_phase: 'Beta')
```
gidsg: as we are repeating these queries now, we could DRY it up by moving them into the register model e.g.
```
  scope :organisation_count, lambda {
                               available
                                 .in_beta
                                 .distinct
                                 .count(:authority)
                             }
  scope :available_count, -> { available.in_beta.count }
```
Then in the template we can call them on the model e.g. `Register.organisation_count`

app/controllers/registers_controller.rb
```diff
@@ -19,6 +19,9 @@ def index
 
   def in_progress
     @registers = Register.available.where(register_phase: 'Alpha').sort_by_phase_name_asc.by_name
+
+    @registers_available = Register.available.where(register_phase: 'Beta')
```
gidsg: I think these instance variables can be safely deleted now (also from `pages_controller.rb`) as we get them directly in the partial now.

injms: As per our discussion I've referenced the model in the controller as a variable passed to the view.

injms:
I'll merge this on Monday morning.

PR #422, opened 2018-09-05
Add theme data

gidsg:
### Context
We need to populate the themes.

Added theme data using a migration as we do not anticipate it changing frequently enough to require an admin UI.

### Guidance to review
Content should match that supplied by @mpwoods


db/migrate/20180905090146_add_theme_data.rb
```diff
@@ -0,0 +1,17 @@
+class AddThemeData < ActiveRecord::Migration[5.2]
+  def change
```
MatMoore: this should be `up` I think, as it just runs when applying the migration, not on the rollback

PR #425, opened 2018-09-06
[DO NOT MERGE] Remove " register" suffix from register names

MatMoore:
### Context
Friendly register names can now contain the " register" suffix, so adding it in the templates generates redundant redundancy.

I found this out the hard way by updating the government service register, which resulted in the frontend displaying "Government service register register".

This is a problem as right now we can't refresh the data in registers-frontend without breaking all the copy on the site.

We think that register frontend shouldn't impose any constraints over the naming of registers - it should be flexible enough to handle whatever we want to call them.

This PR changes all the places where register names are referred to like "the XYZ register" or just "XYZ register" to just "XYZ". This needs content review before merging. In particular, the copy on the updates page doesn't work anymore so I had to change the sentence structure.

### Screenshots
These screenshots show what the website will look like with the current names (some of them will end with " register").

Note that this suffix will not show up when this is first deployed, because production still has old data. The goal of this PR is for registers frontend to work with both naming conventions.

screen shot 2018-09-06 at 17 29 52
screen shot 2018-09-06 at 17 30 27
screen shot 2018-09-06 at 17 31 08

The new names will also be surfaced in page titles and JSON-LD metadata that gets included in the HTML.

### DIFF
--- a/app/views/download/index.html.haml
+++ b/app/views/download/index.html.haml
@@ -1,4 +1,4 @@
-- @page_title = "Download - #{@register.register_name} register"
+- @page_title = "Download - #{@register.register_name}"
 - content_for :body_classes, 'api-key-page'
 
 %main#content{role: 'main'}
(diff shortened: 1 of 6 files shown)


MatMoore:
Going to close this and revert back to the old style names for now. We'd rather not refer to "* register" everywhere. We may still need to modify some of the copy on registers frontend at a later date, because if we start listing registers elsewhere we will lose control over the naming convention, but I think it will be easier to postpone this until we need to deal with it.

PR #430, opened 2018-09-12
Blue bar needs to be (temporarily) removed.

injms:
### Context
The blue bar clashes with the data format change notice - this removes it.

Ternary operator removed, as we just need the `header--without-border` class on every page.

### Guidance to review
None.

### DIFF
--- a/app/views/layouts/application.html.haml
+++ b/app/views/layouts/application.html.haml
@@ -3,9 +3,9 @@
 - content_for :page_title, "#{@page_title.present? ? "#{@page_title} - GOV.UK Registers" : 'GOV.UK Registers' }"
 
 - content_for :after_header do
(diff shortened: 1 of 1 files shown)


app/views/layouts/application.html.haml
```diff
@@ -3,7 +3,8 @@
 - content_for :page_title, "#{@page_title.present? ? "#{@page_title} - GOV.UK Registers" : 'GOV.UK Registers' }"
 
 - content_for :after_header do
-  %header{class: "header #{current_page?(root_path) ? 'header--without-border' : 'header'}"}
+  -# %header{class: "header #{current_page?(root_path) ? 'header--without-border' : 'header'}"}
```
gidsg: Should we add a `TODO` on the comment to restore so it is more easily identified?

PR #435, opened 2018-09-26
Registers collections and frontend changes

injms:
### Context
This is the changes to the registers collections page, the about page, the homepage, and the services using registers page.

 - Register collection page displays categories by default; can show organisations, and alphabetical list
 - About page updated with new copy
 - Homepage updated to show categories
 - Services using registers page updated
 - Footer and navigation updated

### Guidance to review
Preview is on the research staging site attached to each Trello card.


app/assets/javascripts/application.js
```diff
@@ -28,6 +28,8 @@
 // GOV.UK Design System
 //= require all.js
 
+document.getElementsByTagName('html')[0].className += 'js'
```
gidsg: what is this change doing?

injms: By adding a `js` class to the html element, I can hide the submit button on the registers collection page 'Sort by' form using CSS - this means the button doesn't flash in and out of view for those with JavaScript enabled and it shows instantly when people have JavaScript disabled.

MatMoore:
This is probably nothing, but just want to check this is the design.

When you have an empty category, the positioning of the "no registers" copy looks a bit strange to me:
screen shot 2018-09-26 at 11 39 35

I'm assuming this is because the category titles are always in a separate column to the left of the list of registers, rather than being above the list.

app/assets/stylesheets/application.scss
```diff
@@ -649,7 +631,74 @@ details {
...
   padding-left: 0.25em;
 }
+
+.registers-\!-border-off {
+  border: 0;
+}
+
+// ============================================================================
+// Migration hacks
+// ============================================================================
```
MatMoore: will we remove these once (if?) elements is removed?

app/controllers/registers_controller.rb
```diff
@@ -7,14 +7,25 @@ class RegistersController < ApplicationController
...
+    if params[:showby] == 'name' || @search_term.present?
+      @show_by_selected = 'name'
+      @registers_collection = Register.available
+                           .in_beta
+                           .search_registers(@search_term)
+                           .by_name
+    elsif params[:showby] == 'organisation'
+      @show_by_selected = 'organisation'
+      @registers_collection = Authority.with_a_register
+    else
```
MatMoore: Should we hide the themes without a register, as we do for organisations? We are currently showing themes with nothing on them.

app/assets/stylesheets/application.scss
injms: Yes - there's in the next mission for the move to Design System and the removal of Elements.

app/controllers/registers_controller.rb
injms: Good point - I think that every theme/category _should_ have a register assigned to it, but that a assumption on my part.

app/controllers/registers_controller.rb
```diff
@@ -7,14 +7,25 @@ class RegistersController < ApplicationController
...
-    @registers = Register.available
-                         .in_beta
-                         .search_registers(@search_term)
 
     # Redirect legacy URL to ensure we don't break anyone
     if params[:phase] == 'in progress'
       redirect_to registers_in_progress_path
     end
+
+    if params[:showby] == 'name' || @search_term.present?
```
gidsg: can we use `show_by` as it is more consistent with our other params?

injms: Done.

app/controllers/registers_controller.rb
injms: Done and pushed. Cheers!

spec/features/access_data_flow_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    with(headers: { 'Accept' => '*/*', 'Accept-Encoding' => 'gzip, deflate', 'Host' => 'government-organisation.register.gov.uk' }).
+    to_return(status: 200, body: "assert-root-hash\tsha-256:f42d8409df99cceb6c92e5b6b2eb24cd51075d1aa23924fcbdbabf57fbc4ab98")
+
+    country = ObjectsFactory.new.create_register('Country', 'Beta', 'D587')
+    government_organisation = ObjectsFactory.new.create_register('government-organisation', 'Beta', 'D587')
+    PopulateRegisterDataInDbJob.perform_now(country)
+    PopulateRegisterDataInDbJob.perform_now(government_organisation)
+  end
+
+  # scenario 'search for a register' do
```
MatMoore: is this meant to be commented out?

spec/features/access_data_flow_spec.rb
```diff
@@ -0,0 +1,44 @@
+require 'rails_helper'
+
+RSpec.feature 'View register', type: :feature do
+  before(:all) do
+    country_data = File.read('./spec/support/country.rsf')
```
MatMoore: I haven't thought about this in detail, but do you think we could improve some of this set up and stubbing in the feature tests?

It seems like some of this setup is common to multiple tests, and if the tests aren't actually testing the consumer process that reads data in from ORJ there would be easier ways to get to a state where we have a few registers in the DB and then you can just focus on testing the user journey.

spec/controllers/category_controller_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+
+RSpec.describe CategoryController, type: :controller do
+  describe "GET #index" do
+    it "returns http success" do
+      get :index
+      expect(response).to have_http_status(301)
+    end
+  end
+
+  # describe "GET #show" do
```
MatMoore: is this meant to be commented out?

spec/controllers/category_controller_spec.rb
```diff
@@ -0,0 +1,17 @@
+require 'rails_helper'
+
+RSpec.describe CategoryController, type: :controller do
+  describe "GET #index" do
+    it "returns http success" do
```
MatMoore: the description doesn't match the expectation here. If it's a 301 maybe we can assert that it redirects to a particular page.

spec/controllers/authority_controller_spec.rb
```diff
@@ -0,0 +1,17 @@
+require 'rails_helper'
+
+RSpec.describe AuthorityController, type: :controller do
+  describe "GET #index" do
+    it "returns http success" do
```
MatMoore: see below- the description doesn't match the expectation

app/views/registers/index.html.haml
```diff
@@ -1,85 +1,125 @@
...
-                - @registers.featured.by_popularity.each do |register|
-                  %li.app-c-highlight-boxes__item-wrapper
-                    .app-c-highlight-boxes__item
-                      %h3.app-c-highlight-boxes__title= link_to register.register_name, register_path(register.slug)
-                      %p.app-c-highlight-boxes__description= register.register_description
-                      - if register.authority.present?
-                        %div{class: "govuk-organisation-logo #{register&.authority&.name.parameterize}"}
-                          %div{class: "logo-container #{crest_class_name(register&.authority&.name.parameterize)}"}
-                            %span= register&.authority&.name
+    .govuk-grid-row#show-by
```
MatMoore: Is any of this common to the category pages? If so maybe we can extract a partial here to avoid duplication.

app/assets/javascripts/application.js
MatMoore: Can you add a comment for it? I think it makes sense but I wouldn't have understood it without that explanation.

injms: ✍️ done

injms:
> This is probably nothing, but just want to check this is the design.
> 
> When you have an empty category, the positioning of the "no registers" copy looks a bit strange to me:
> screen shot 2018-09-26 at 11 39 35
> 
> I'm assuming this is because the category titles are always in a separate column to the left of the list of registers, rather than being above the list.

Good call! Now that the category doesn't appear on either the homepage or the register collection page, the only way to see a category that doesn't have any registers in it is to guess the address - I think it's best to have a page explaining why rather than a 404. @conordelahunty said he's happy with the layout with a change of copy, which I've now done.

spec/controllers/category_controller_spec.rb
injms: I've edited this so that it tests for both a 301, and that the redirect ends up on the registers collection page.

spec/controllers/authority_controller_spec.rb
injms: I've also edited this so it tests for both a 301, and that it redirects to the registers collection page, showing by organisations.

spec/features/access_data_flow_spec.rb
injms: I agree. I think this is too big for this PR, so I'll raise a ticket that we need to look at the testing set up  (and rolled into this, also the current difficulty doing any browser-based JavaScript testing.)

spec/controllers/category_controller_spec.rb
injms: This test has been removed, since I agree that we need to improve the test set up since so many tests now rely on a common set up (as in your other comment).

spec/features/access_data_flow_spec.rb
injms: You're correct - this should be commented out. It's debt from a previous piece of work that is now part of the 'look at testing' ticket. I've started writing the tests, but have commented this out for this branch.

app/views/registers/index.html.haml
injms: I've put a list partial on both the categories and the organisations view on this template.

app/assets/stylesheets/modules/_register-categories.scss
```diff
@@ -0,0 +1,167 @@
...
+  margin: 0;
+  padding: 0.75em 0.75em 0.6em 0.75em;
+  text-decoration: none;
+
+  &:hover {
+    border: 1px solid $govuk-link-colour;
+    box-shadow: 0 0 0 1px $govuk-link-colour;
+  }
+
+  // @media (hover) {
```
gidsg: can we delete this?

gidsg:
when I try to go to a category page, I am getting
```
ActionView::MissingTemplate in Category#show
Missing partial layouts/_list-of-registers
```
Are you sure all files are committed?

gidsg:
I get this warning in the console when rendering the homepage:
```
Creating scope :organisation_count. Overwriting existing method Register.organisation_count.
```

app/models/authority.rb
```diff
@@ -1,3 +1,21 @@
...
   has_many :registers
+
+  scope :by_name, -> { order name: :asc }
+
+  scope :with_a_register, -> {
+    joins(:registers).merge(Register.in_beta).distinct.by_name
+  }
+
+  scope :collection, ->(id) {
+    Authority.find_by(government_organisation_key: id)
```
gidsg: We should use [find_by!](https://api.rubyonrails.org/classes/ActiveRecord/FinderMethods.html#method-i-find_by-21) here as we want to 404 if not found rather than 500.

app/assets/stylesheets/modules/_register-categories.scss
injms: We can.

app/models/authority.rb
injms: Changes in both the authority and category models.

app/views/layouts/_themes.html.haml
```diff
@@ -0,0 +1,14 @@
+%ul.register-themes
+  - @registers_collection.each do |category|
+    - if category.slug == 'digital-data-and-technology-profession-capability-framework' and current_page?(root_path)
```
gidsg: can we move this out of the template and into the controller or model?

gidsg: I think we could query: `Category.where.not(slug: 'digital-data-and-technology-profession-capability-framework')`

injms: Logic now resides in model - can you give it a check to see if you're happy with it?

app/models/category.rb
```diff
@@ -7,6 +7,12 @@ class Category < ApplicationRecord
     joins(:registers).merge(Register.in_beta).distinct.by_name
   }
 
+  scope :with_a_register__homepage, -> {
+    Category.with_a_register.reject{ |r|
```
gidsg: I think we could use `where.not(slug: 'digital-data-and-technology-profession-capability-framework')`

app/models/category.rb
```diff
@@ -7,6 +7,12 @@ class Category < ApplicationRecord
     joins(:registers).merge(Register.in_beta).distinct.by_name
   }
 
+  scope :with_a_register__homepage, -> {
```
gidsg: would `shown_on_homepage` be a clearer name?

app/models/application_record.rb
```diff
@@ -1,3 +1,7 @@
 class ApplicationRecord < ActiveRecord::Base
   self.abstract_class = true
+
+  scope :registers_available, -> {
```
gidsg: I don't think we need this in `application_record` can it just be in the registers model if at all?

PR #437, opened 2018-10-01
fix rake authorities perform_now rake task

gidsg:
### Context
Rake task was not previously appearing in `rake --tasks`

Fix rake task syntax

### Guidance to review
`rake registers_frontend:populate_authorities:fetch_now` should appear in `rake --tasks`

### DIFF
--- a/lib/tasks/populate_authorities.rake
+++ b/lib/tasks/populate_authorities.rake
@@ -1,9 +1,11 @@
 namespace :registers_frontend do
   namespace :populate_authorities do
-    desc "Populate Authorities using Government Organisation register."
+    desc "Add task to queue: Populate Authorities using Government Organisation register."
     task fetch_later: :environment do
       PopulateAuthoritiesJob.perform_later
     end
+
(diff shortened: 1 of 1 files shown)


injms:
> On my local install, `theme_id` throws an error, as we changed that to `category_id`. I just wanted to check that the `theme_id` in the admin/registers view is magically changed to `category_id`? Or is that a manual thing I need to remember?
> 
> [registers-frontend/app/views/admin/registers/_form.html.haml](https://github.com/openregister/registers-frontend/blob/9e7745f0a42d07cf6a726d95938363474c4a0908/app/views/admin/registers/_form.html.haml#L12-L13)
> Lines 12 to 13 in [9e7745f](/openregister/registers-frontend/commit/9e7745f0a42d07cf6a726d95938363474c4a0908)
>  = f.label :theme_id, class: 'form-label' 
>  = f.select :theme_id, Theme.all.pluck(:name, :id), { include_blank: 'Select theme' }, class: 'form-control'

Ignore this - this would've happened if any branch had been used locally.

PR #440, opened 2018-10-03
Uses the correct name when listing registers.

injms:
### Context
Registers were being listed with the wrong name - this pull request corrects that.

When listed, the register now uses `register_name` rather than `name` - this gives the more human readable name.

### Guidance to review
None.


injms:
Backlog card created - cheers!

PR #459, opened 2018-10-10
Prevents records with a future end date being displayed as archived.

injms:
### Context
Any record that had any end date was filtered into the archive - whether the date was in the future or not. This isn't the desired behaviour.

* Changes the archive filter to only include past end dates
* Changes the current filter to include future end dates and no end dates

### Guidance to review
Line 11 and 12 contain two queries (one commented out) that give the same displayed results. Which of these is preferable?

### DIFF
--- a/app/models/record.rb
+++ b/app/models/record.rb
@@ -3,8 +3,15 @@ class Record < ApplicationRecord
 
   include SearchScope
   belongs_to :register
-  scope :current, -> { where(Arel.sql("data->> 'end-date' is null")) }
-  scope :archived, -> { where(Arel.sql("data->> 'end-date' is not null")) }
+  scope :current, -> {
+    where(Arel.sql("data->> 'end-date' is null or data->> 'end-date' > '#{Date.today}'"))
+  }
+
+  scope :archived, -> {
+    # where("data->> 'end-date' < ?", Date.today)
+    where(Arel.sql("data->> 'end-date' < '#{Date.today}'"))
+  }
+
   scope :status, lambda { |status|
     case status
     when 'archived', 'current'
--- a/app/models/record.rb
+++ b/app/models/record.rb
@@ -4,7 +4,7 @@ class Record < ApplicationRecord
   include SearchScope
   belongs_to :register
   scope :current, -> {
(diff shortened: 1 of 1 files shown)


app/models/record.rb
```diff
@@ -3,8 +3,15 @@ class Record < ApplicationRecord
...
   include SearchScope
   belongs_to :register
-  scope :current, -> { where(Arel.sql("data->> 'end-date' is null")) }
-  scope :archived, -> { where(Arel.sql("data->> 'end-date' is not null")) }
+  scope :current, -> {
+    where(Arel.sql("data->> 'end-date' is null or data->> 'end-date' >= '#{Date.today}'"))
+  }
+
+  scope :archived, -> {
+    # where("data->> 'end-date' < ?", Date.today)
```
gidsg: I think this version is preferable rather than interpolating within the SQL.

gidsg: This field is actually denormalised directly on the record so we should be able to do:
`where('end_date < ?', Date.today)` with equivalence

app/models/record.rb
```diff
@@ -3,8 +3,15 @@ class Record < ApplicationRecord
 
   include SearchScope
   belongs_to :register
-  scope :current, -> { where(Arel.sql("data->> 'end-date' is null")) }
-  scope :archived, -> { where(Arel.sql("data->> 'end-date' is not null")) }
+  scope :current, -> {
+    where(Arel.sql("data->> 'end-date' is null or data->> 'end-date' >= '#{Date.today}'"))
```
gidsg: I think this should be `>` rather than `>=` as would would consider one with an end date of today to be archived.

app/models/record.rb
injms: Super - I've switched to the Rails magic version. There are another couple of queries using `Arel.sql` in this model - but I think that's a non-urgent fix for another day.

app/models/record.rb
injms: Okay - I've put archived to be `<=` instead of `<` so that a day won't be missed two queries.

app/models/record.rb
gidsg: Yes, we have to use `Arel.sql` to query JSONB as it is not part of standard SQL.

PR #465, opened 2018-10-15
Preview the data changes

injms:
### Context
This updates the individual register page layout to meet the [requirements in the ticket](https://trello.com/c/Hf68Vi4C/95-implement-mvp-design-of-preview-the-data).

 * Register show template partially updated and now uses Design System (both JavaScript and non-JavaScript versions)
 * Limits the number of records displayed to 10; paginated with no-JS, loads 10 more at a time with JS.
 * Records model updated to remove archived and current filter
 * Event tracking added

### Guidance to review
 * The accordion is a tweaked version of the old accordion; the reason this hasn't moved across to the new Design System is that there isn't an accordion in it...[yet](https://github.com/alphagov/govuk-frontend/pull/958).

### DIFF
--- a/app/assets/javascripts/accordion-with-descriptions.js
+++ b/app/assets/javascripts/accordion-with-descriptions.js
@@ -141,8 +141,11 @@
       function bindToggleForSubsections() {
         // Add toggle functionality individual sections
         $subsectionHeader.on('click', function(e) {
-          toggleSection($(this).next());
-          toggleIcon($(this));
+          $(this).parent().find('.subsection__content').each( function() {
+            toggleSection( $(this) )
+            toggleIcon($(this));
+        })
+
           toggleState($(this).find('.subsection__button'));
           //setOpenCloseAllText();
           if(window.ga && ga.create) {
--- a/app/assets/javascripts/accordion-with-descriptions.js
+++ b/app/assets/javascripts/accordion-with-descriptions.js
@@ -16,7 +16,7 @@
       var totalSubsections = $element.find('.subsection__content').length;
 
       var $openOrCloseAllButton;
-      var GOVUKServiceManualTopic = serviceManualTopicPrefix();
+      var RegisterAccordionSection = serviceManualTopicPrefix();
 
       //addOpenCloseAllButton();
       addButtonsToSubsections();
@@ -108,7 +108,7 @@
 
         $subsectionContent.each(function(index) {
           var subsectionContentId = $(this).attr('id');
-          if(sessionStorage.getItem(GOVUKServiceManualTopic+subsectionContentId)){
+          if(sessionStorage.getItem(RegisterAccordionSection + subsectionContentId)){
             openStoredSections($("#"+subsectionContentId));
           }
         });
@@ -124,7 +124,7 @@
             $(this)
               .find('.subsection__content')
               .each(function(i) {
-                sessionStorage.setItem( GOVUKServiceManualTopic + $(this).attr('id') , 'Opened');
+                sessionStorage.setItem( RegisterAccordionSection + $(this).attr('id') , 'Opened');
               })
           })
         }
@@ -139,7 +139,7 @@
               .find('.subsection__content')
               .each(function(i) {
                 var subsectionClosedContentId = $(this).attr('id')
-                sessionStorage.removeItem( GOVUKServiceManualTopic + subsectionClosedContentId , subsectionClosedContentId);
+                sessionStorage.removeItem( RegisterAccordionSection + subsectionClosedContentId , subsectionClosedContentId);
               })
           })
         }
--- a/app/assets/javascripts/accordion-with-descriptions.js
+++ b/app/assets/javascripts/accordion-with-descriptions.js
@@ -51,12 +51,23 @@
       }
 
       function addButtonsToSubsections() {
-        var $subsectionTitle = $element.find('.subsection__title');
+        var $subsectionTitle = $element.find('.subsection__title')
 
-        // Wrap each title in a button, with aria controls matching the ID of the subsection
-        $subsectionTitle.each(function(index) {
-          $(this).wrapInner( '<button class="subsection__button" aria-expanded="false" aria-controls="subsection_content_' + index +'"></a>' );
-        });
+        // Wrap each title in a button, with aria controls matching the ID(s) of the subsection
+        $subsectionTitle.each( function(index) {
+          var $subsections = $(this)
+                                .closest('.js-accordion-with-descriptions')
+                                  .find('.js-accordion-subsection')
+
+          var subsectionsIds = [];
+
+          for (var i = 0; i < $subsections.length; i++) {
+            subsectionsIds.push($subsections[i].id)
+          }
+
+          $(this)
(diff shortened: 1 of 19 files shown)


app/assets/javascripts/accordion-with-descriptions.js
```diff
@@ -121,9 +121,12 @@
         if (isOpenSubsections) {
           var $openSubsections = $('.subsection--is-open');
           $openSubsections.each(function(index) {
-            var subsectionOpenContentId = $(this).find('.subsection__content').attr('id');
-            sessionStorage.setItem( GOVUKServiceManualTopic+subsectionOpenContentId , 'Opened');
-          });
+            $(this)
+              .find('.subsection__content')
+              .each(function(i) {
+                sessionStorage.setItem( GOVUKServiceManualTopic + $(this).attr('id') , 'Opened');
```
gidsg: Is `GOVUKServiceManualTopic` some boiler plate, can we change it to `RegisterPageSection` or whatever it represents?

app/models/register.rb
```diff
@@ -30,6 +30,9 @@ class Register < ApplicationRecord
       .count(:authority_id)
   }
   scope :available_count, -> { available.in_beta.count }
+  scope :category, ->(category_id) {
```
gidsg: we don't need this, we can use rails associations instead. See: https://github.com/openregister/registers-frontend/blob/6efb3e596f25a77ac47a4600d61553457ddf2880/app/models/register.rb#L40 if we fix this association and change it to `:category` then you can just call through Registers model for the category e.g. `Register.first.category`. This also negates the need for the  instance variable in the controller.

app/views/registers/show.html.haml
```diff
@@ -2,142 +2,139 @@
...
         %p.govuk-body-l= @register.register_description
-        - unless @register.is_empty?
-          %p.font-xsmall Last updated: #{link_to formatted_date(@register.register_last_updated), register_entries_path(@register.slug)}
-
-      .column-one-third
+        %dl.govuk-metadata
+          - if @register.authority.present?
+            %dt.register-metadata__term Organisation:
+            %dd.register-metadata__definition= link_to @register&.authority&.name, authority_path(@register.authority.government_organisation_key)
+          - if @register&.category_id&.present? or @register.register_phase != 'Beta'
```
gidsg: we could move this logic into the model e.g.
```
 def show_category_link
   register_phase == 'Beta' && category.present?
 end
```

app/views/registers/show.html.haml
```diff
@@ -2,142 +2,139 @@
...
+          %div{class: "govuk-organisation-logo #{@register&.authority&.name&.parameterize}"}
+            %div{class: "logo-container #{crest_class_name(@register&.authority&.name&.parameterize)}"}
+              %span= @register&.authority&.name
+    %section.govuk-grid-row#records_wrapper{ data: { module: 'accordion-with-descriptions'}}
+      .govuk-grid-column-full.subsection__header#search_wrapper
+        %h2.govuk-heading-m.subsection__title
+          About the data
+      .govuk-grid-column-two-thirds.subsection__content#about-the-data__1
+        %p This register has #{number_with_delimiter(@register_records_total_count)} records and #{@register.register_fields.count} fields
+        -# = page_records_info(@records).inspect
```
gidsg: we should remove this!

app/views/registers/show.html.haml
```diff
@@ -2,142 +2,139 @@
...
-              %h2.subsection__title Give feedback
-            .subsection__content.js-subsection-content#feedback_section
-              = render partial: 'feedback/form'
+      .govuk-grid-column-full.subsection__content#about-the-data__5
+        %p= link_to 'Access all the data in this register', register_choose_access_path(@register.slug), { id: 'access-the-data-from-preview', class: 'registers__button--access-large'}
+        - unless @records.any? || @register.is_empty?
+          - if @search_term.present?
+            %script
+              -# TODO: current and archived no longer present; what goes here as the filter?
+              var searchFilter = 'current'; //
```
gidsg: If we are removing the filtering options I guess we should remove this extra tracking logic too?

app/views/registers/show.html.haml
injms: 👍

app/views/registers/show.html.haml
injms: Fixed.

app/assets/javascripts/accordion-with-descriptions.js
injms: I've changed the variable names - but I want to avoid changing the accordion javascript any more. There is other work that could be done on it to make the accordion more effective, but I want to know that an accordion will be used in the future and to see what the new accordion does before jumping into this in any more depth.

app/models/register.rb
injms: Sure thing - I've changed it to this method.

app/views/registers/show.html.haml
injms: Sounds good; now done.

spec/features/view_register_spec.rb
```diff
@@ -82,12 +74,12 @@
     expect(first_update.sibling('table')).to have_content("The Republic of Cote D'Ivoire")
   end
 
-  scenario 'view register name' do
-    visit('/registers/industrial-classification-2003')
-    register_name = 'UK Standard Industrial Classification of Economic Activities 2003 register'
-    expect(find('h1')).to have_content(register_name)
-    expect(page.title).to have_content(register_name)
-  end
+  # scenario 'view register name' do
```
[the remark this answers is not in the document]
injms: 👍

app/models/register.rb
```diff
@@ -30,11 +30,12 @@ class Register < ApplicationRecord
       .count(:authority_id)
   }
   scope :available_count, -> { available.in_beta.count }
+  scope :category, -> { first.category }
```
gidsg: Is this needed, looks like it would work the same as that provided by the model association?

injms: It's now gone.

app/views/registers/show.js.erb
```diff
@@ -1,8 +1,10 @@
-$("#records-tbody").append("<%= j(render(partial: 'record', collection: @records)) %>");
-$("#records-count").html('Showing <strong>1 &ndash; <%= @records.count + @records.offset_value %></strong> of <strong><%= @records.total_count %></strong> records <% if @search_term.present? %>for <strong>"<%= search_term %>"</strong><% end %>')
+document.getElementById('records-tbody').innerHTML += "<%= j(render(partial: 'record', collection: @records)) %>";
+
+document.getElementById('records-count').innerHTML = 'Showing <strong>1 &ndash; <%= @records.count + @records.offset_value %></strong> of <strong><%= @records.total_count %></strong> records <% if @search_term.present? %>for &ldquo;<strong><%= @search_term %></strong>&rdquo;<% end %>';
```
gidsg: should we use `number_with_delimiter` here too for consistency?

injms: Oops. Now committed.

PR #471, opened 2018-10-23
Removes table sorting

injms:
### Context
Removes the front-end functionality that allows each table to be sorted by the column; defaults to 'Name', then the first column.

* Moves the logic and the elements into the helper to avoid logic in the templates
* Style changes to account for the change in markup.

### Guidance to review
None.


gidsg:
...also `:sort_by_field` scope and its usages can be deleted.

injms:
> ...also `:sort_by_field` scope and its usages can be deleted.

I tidied up the controller very slightly - but removing the `:sort_by_field` scope and its use in the controller causes the records to be displayed unsorted, so I've left that functionality in.

PR #474, opened 2018-10-24
Questionnaire form changes

injms:
### Context

The form needs to have certain fields made mandatory.

The initial question - _Do you work for the government?_ - is now required. If answering &lsquo;Yes&rsquo; to this question, then a further two questions are mandatory:
 * Which part of government are you from?
 * What are you using registers for?

* Makes the following question required:
  * Do you work for the government?
* Makes the following questions required if answered &lsquo;Yes&rsquo; to working in government:
  * Which part of government are you from?
  * What are you using registers for?
* Shows error messages when required fields have not been filled in

### Guidance to review
The questions are still skippable by clicking either the &lsquo;Skip this step&rsquo; or the &lsquo;I&apos;d prefer not to say&rsquo; step.

### DIFF
--- a/app/views/download/help_improve.html.haml
+++ b/app/views/download/help_improve.html.haml
@@ -16,12 +16,12 @@
 = content_for :javascript do
   :javascript
     (function(document, window, undefined) {
-      function isIE () {
-        var myNav = navigator.userAgent.toLowerCase();
-        return (myNav.indexOf('msie') != -1) ? parseInt(myNav.split('msie')[1]) : false;
+      var isIE = function() {
+        var myNav = navigator.userAgent.toLowerCase()
+        return (myNav.indexOf('msie') != -1) ? parseInt(myNav.split('msie')[1]) : false
       }
 
-      if (isIE() && isIE() >= 9) {
+      if (isIE() && isIE() >= 9 || isIE() === false ) {
         var forms = document.querySelectorAll('.js--ga-submit')
 
         var sendData = function(form) {
@@ -144,10 +144,8 @@
           var messageTarget = form.querySelector(info.messageTarget)
           var messageText = document.createTextNode(info.messageText)
 
-          var message = document.createElement('span');
-          message.className = 'govuk-error-message'
-          message.appendChild(messageText)
-          messageTarget.parentNode.insertBefore(message, messageTarget.nextSibling)
+          messageTarget.className = 'govuk-error-message'
+          messageTarget.appendChild(messageText)
 
           classTarget.className += ' govuk-form-group--error'
         }
@@ -178,7 +176,7 @@
               if (partOfGov(form) === false) {
                 showWarning(form, {
                   classTarget: '.js--part-of-gov-group',
-                  messageTarget: '.js--part-of-gov-label',
+                  messageTarget: '.js--part-of-gov-error-message',
                   messageText: 'Enter which part of government you’re from'
                 })
               }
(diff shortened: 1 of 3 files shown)


app/views/download/help_improve.html.haml
```diff
@@ -16,41 +16,30 @@
...
-          return false;
+      var isIE = function() {
+        var myNav = navigator.userAgent.toLowerCase()
+        return (myNav.indexOf('msie') != -1) ? parseInt(myNav.split('msie')[1]) : false
       }
 
-      for (var i = 0; i < forms.length; i += 1) {
-        forms[i].addEventListener('submit', function(event) {
-          var form = event.target
+      if (isIE() && isIE() >= 9 || isIE() === false ) {
```
MatMoore: What's the functionality that's not available in old IE? And can we change this to check whether that's available, instead of parsing the user agent?

app/views/download/help_improve.html.haml
```diff
@@ -16,41 +16,30 @@
...
+        var sendData = function(form) {
+
+          var getWhatRegistersAreBeingUsedFor = function(form) {
+            var isGovernment = isGov(form)
+            if (isGovernment === true) {
+              return getElementValue(form.querySelector('input[name="help_us_improve_user[gov_what_are_you_using_registers_for]"]:checked'))
             }
-            else if (isGovernment === "No") {
-              return getValue(form.querySelector('input[name="help_us_improve_user[nongov_what_are_you_using_registers_for]"]:checked'))
+            else if (isGovernment === false) {
```
MatMoore: Can this be simplified to just if/else?

app/views/download/help_improve.html.haml
```diff
@@ -16,41 +16,30 @@
...
+      if (isIE() && isIE() >= 9 || isIE() === false ) {
+        var forms = document.querySelectorAll('.js--ga-submit')
 
-          var isGovernment = getValue(form.querySelector('input[name="help_us_improve_user[is_government]"]:checked'))
-          var whatAreYouUsingRegistersFor = function() {
-            if (isGovernment === "Yes") {
-              return getValue(form.querySelector('input[name="help_us_improve_user[gov_what_are_you_using_registers_for]"]:checked'))
+        var sendData = function(form) {
+
+          var getWhatRegistersAreBeingUsedFor = function(form) {
```
MatMoore: Since we've got a lot of code here to extract data from the form I'm wondering if it makes sense to have an object with all these functions as methods. I think it would help to separate out some of the different concerns (parsing the form, talking to GA, and presentational stuff).

app/views/download/help_improve.html.haml
```diff
@@ -85,22 +74,149 @@
             eventLabel: 'Help us improve',
             nonInteraction: true
           })
+        }
 
+        var getElementValue = function(element) {
```
MatMoore: Can we use JQuery for this? Or are we removing that as a dependency?

injms: The Design System has just removed jQuery as a dependency and Rails 5.1 has removed it from the default stack (I think you can still choose for it to be included in Rails). Whilst we still use it on the current frontend, I've avoided it here to try and avoid creating some technical debt.

app/views/download/help_improve.html.haml
injms: Agreed - I've done this in this [different branch](https://github.com/openregister/registers-frontend/pull/481), since a ticket was raised to make the entire form mandatory.

injms:
I'm closing this pull request, as [pull request 481](https://github.com/openregister/registers-frontend/pull/481) supersedes it.

PR #479, opened 2018-10-29
Suggest a register

injms:
### Context
There needs to be a process for new registers to be suggested. This is the initial method, as per the [ticket](https://trello.com/c/1Lixtcme/2980-implement-create-a-register-front-end-flow-3) raised.

 * New page with a form to check that people suggesting a register fulfil certain criteria
    * Autocomplete lists government departments and councils / local authorities in England, Northern Ireland, Wales, and Scotland.
    * Error messaging
 * Thank you page, linking to Google Form with further questions
 * Link from About page to Suggest a register page
 * GA tracking in place

### Guidance to review
 * Merging of the local authority and government department registers was done with a helper that was already in place. I think this can be sped up and am looking into different methods.

### DIFF
--- a/app/controllers/suggest_a_register_controller.rb
+++ b/app/controllers/suggest_a_register_controller.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+class SuggestARegisterController < ApplicationController
+  invisible_captcha only: :create_feedback, honeypot: :spam
+
+  def index
+    @government_organisations = Register.find_by(slug: 'local-authority-eng')
+                                        .records
+                                        .where(entry_type: 'user')
+                                        .current
+  end
+
+  def check
+    if params['suggest_a_register']['gov_what_part_of_government'].present?
+      redirect_to suggest_a_register_complete_path
+    else
+      flash[:summary_title] = 'There is a problem'
+      flash[:summary_message] = 'A government department needs to be selected'
+      flash[:error_message] = 'Select a government department'
+      redirect_to suggest_a_register_path
+    end
+  end
+
+  def complete; end
+end
--- a/app/controllers/suggest_a_register_controller.rb
+++ b/app/controllers/suggest_a_register_controller.rb
@@ -1,13 +1,12 @@
 # frozen_string_literal: true
 
 class SuggestARegisterController < ApplicationController
+  include FormHelpers
+  helper_method :government_orgs_local_authorities
   invisible_captcha only: :create_feedback, honeypot: :spam
 
   def index
-    @government_organisations = Register.find_by(slug: 'local-authority-eng')
-                                        .records
-                                        .where(entry_type: 'user')
-                                        .current
+    @government_organisations = government_orgs_local_authorities
   end
 
   def check
--- a/app/controllers/suggest_a_register_controller.rb
+++ b/app/controllers/suggest_a_register_controller.rb
@@ -11,12 +11,12 @@ class SuggestARegisterController < ApplicationController
 
   def check
     if params['suggest_a_register']['gov_what_part_of_government'].present?
-      redirect_to suggest_a_register_complete_path
+      redirect_to suggest_new_register_complete_path
     else
       flash[:summary_title] = 'There is a problem'
       flash[:summary_message] = 'A government department needs to be selected'
       flash[:error_message] = 'Select a government department'
-      redirect_to suggest_a_register_path
+      redirect_to suggest_new_register_path
     end
   end
 
--- a/app/controllers/suggest_a_register_controller.rb
+++ b/app/controllers/suggest_a_register_controller.rb
@@ -13,9 +13,7 @@ class SuggestARegisterController < ApplicationController
     if params['suggest_a_register']['gov_what_part_of_government'].present?
       redirect_to suggest_new_register_complete_path
     else
-      flash[:summary_title] = 'There is a problem'
-      flash[:summary_message] = 'A government department needs to be selected'
-      flash[:error_message] = 'Select a government department'
+      flash[:error_message] = 'Enter a government or public sector organisation'
       redirect_to suggest_new_register_path
     end
   end
--- a/app/views/suggest_a_register/complete.haml
+++ b/app/views/suggest_a_register/complete.haml
@@ -1,0 +1,28 @@
+- @page_title = 'Suggest a register'
+- content_for :body_classes, 'suggest-a-register'
+
+%nav.breadcrumbs{role: "Navigation", "aria-label" => "Breadcrumb"}
+  %ol{data: {"click-events" => "", "click-category" => "Header", "click-action" => "Breadcrumb Clicked"}}
+    %li.breadcrumbs__item
+      = link_to 'GOV.UK Registers', root_path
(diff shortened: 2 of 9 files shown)


app/controllers/suggest_a_register_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+  invisible_captcha only: :create_feedback, honeypot: :spam
+
+  def index
+    @government_organisations = government_orgs_local_authorities
+  end
+
+  def check
+    if params['suggest_a_register']['gov_what_part_of_government'].present?
+      redirect_to suggest_a_register_complete_path
+    else
```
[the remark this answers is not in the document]
MatMoore: Another potential problem with having a separate step to collect the organisation is there seems to be no way to link up the data about organisations with the actual form responses?

app/views/suggest_a_register/complete.haml
```diff
@@ -0,0 +1,47 @@
...
+      .govuk-grid-column-full
+        %h1{class: 'govuk-heading-l govuk-!-margin-bottom-3'} Suggest a new register
+      .govuk-grid-column-two-thirds
+        %p You need to complete our engagement form to suggest a new register.
+        %p The details you provide will help us to better understand your needs for a register, the data you have to offer and whether it is suitable to become a register.
+        %p.govuk-inset-text
+          = link_to 'Use our Google form', 'https://docs.google.com/forms/d/e/1FAIpQLSfw54dI5uYuF-15zH6Iu7KRQnj8lRY3Wcl9JI5hC80cENx-jw/viewform', {data: {'ga-label': 'Suggest a register Google form'}}
+          &nbsp;to suggest a new register.
+
+        %h3{class: 'govuk-heading-s govuk-!-margin-bottom-1'} Can’t access Google forms?
```
MatMoore: Do we really have to accommodate policy that blocks civil servants from using basic office tools? 😩

app/views/suggest_a_register/complete.haml
```diff
@@ -0,0 +1,47 @@
...
+        %p
+          = link_to 'Contact the GOV.UK Registers team', support_question_url
+          &nbsp;for support
+
+
+= content_for :javascript do
```
MatMoore: This code seems reusable, can we create a shared function or something?

app/views/suggest_a_register/index.html.haml
```diff
@@ -0,0 +1,68 @@
...
+              What part of government are you from?
+            - if flash[:error_message].present?
+              %span{id: 'what-part-of-gov-error', class: 'js--what-part-of-gov-error govuk-error-message govuk-!-margin-top-2'}= flash[:error_message]
+            = select( 'suggest_a_register', 'gov_what_part_of_government', @government_organisations.map { |org| [org[1], org[0]]}, { include_blank: true }, id: 'government-organisations' )
+          %button{ class: 'govuk-button', type: 'submit'} Continue
+
+
+= content_for :javascript do
+  :javascript
+    accessibleAutocomplete.enhanceSelectElement({
```
MatMoore: If we do use this widget, you can configure it to support synonyms as well as the actual org names. I don't know if it's worth the effort for what we want to achieve, but we could support common acronyms like HMRC MOJ DWP etc, as these don't work atm. If you speak to GOV.UK they should be able to give you a list of these.

db/migrate/20181026104718_create_suggest_a_registers.rb
```diff
@@ -0,0 +1,7 @@
+class CreateSuggestARegisters < ActiveRecord::Migration[5.2]
+#   def change
```
MatMoore: Is this still needed?

app/views/suggest_a_register/complete.haml
```diff
@@ -0,0 +1,47 @@
...
+      .govuk-grid-column-two-thirds
+        %p You need to complete our engagement form to suggest a new register.
+        %p The details you provide will help us to better understand your needs for a register, the data you have to offer and whether it is suitable to become a register.
+        %p.govuk-inset-text
+          = link_to 'Use our Google form', 'https://docs.google.com/forms/d/e/1FAIpQLSfw54dI5uYuF-15zH6Iu7KRQnj8lRY3Wcl9JI5hC80cENx-jw/viewform', {data: {'ga-label': 'Suggest a register Google form'}}
+          &nbsp;to suggest a new register.
+
+        %h3{class: 'govuk-heading-s govuk-!-margin-bottom-1'} Can’t access Google forms?
+        %p
+          = link_to 'Contact the GOV.UK Registers team', support_question_url
```
MatMoore: I assume this is an existing form that goes to zendesk?

db/schema.rb
```diff
@@ -97,6 +97,11 @@
     t.bigint "authority_id"
   end
 
+  create_table "suggest_a_registers", force: :cascade do |t|
```
MatMoore: This contradicts the commented out migration

app/views/suggest_a_register/complete.haml
injms: Blocking policy aside, this is the first draft of this flow - if it's used a lot, fingers crossed, then we can put more effort into making a non-Google form that everyone can use.

app/views/suggest_a_register/complete.haml
injms: There's only need to track external links on this page at the moment - but yes, it could be pushed site wide. I want to check with Cliff first, since I don't want to swamp Analytics with data that isn't required.

app/views/suggest_a_register/index.html.haml
injms: Ooh, I didn't know that. I'll add the acronyms in.

app/controllers/suggest_a_register_controller.rb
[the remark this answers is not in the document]
injms: The spec asked for this question to be mandatory - the reasoning is that we can only help and host government registers. (The definition of government being the interesting thing.)

A yes / no answer would serve the same purpose, allow people to use their interpretation of government when answering, and avoid duplication with the form later. I think that's a conversation worth having.

We could parse email addresses, but I'd rather not since we then run into the duplication in the main form again. It's also brittle, and we then run the risk of excluding organisations with non-`.gov.uk` email addresses, such as the NHS.

app/views/suggest_a_register/complete.haml
injms: Yep, it goes to the [question form](https://www.registers.service.gov.uk/support/question).

db/migrate/20181026104718_create_suggest_a_registers.rb
injms: Doesn't look like it - I've removed it and seems to deploy okay on staging.

db/schema.rb
injms: I've removed both.

app/views/suggest_a_register/index.html.haml
injms: This needs a slightly different accessible-autocomplete widget - but it's still an improvement we can make. I'll write a ticket to do this and put it in the queue.

PR #481, opened 2018-10-31
Ensures all questions are answered before submitting.

injms:
### Context
The access to data questionnaire needs all questions to be answered before submitting.

* Makes all questions required
* Verifies only the questions relevant - only checks the government questions when the visitor has said they work for government, and only checks the non-government question when the visitor has said that they don't work for government.
* Shows error messages when required fields have not been filled in

### Guidance to review
The questions are still skippable by clicking either the ‘Skip this step’ or the ‘I'd prefer not to say’ step.

### DIFF
--- a/app/views/download/help_improve.html.haml
+++ b/app/views/download/help_improve.html.haml
@@ -19,7 +19,8 @@
       var forms = document.querySelectorAll('.js--ga-submit')
 
       var state = {
-        errorSummary: []
+        errorSummary: [],
+        documentTitle: document.title
       }
 
       var readFormFn = {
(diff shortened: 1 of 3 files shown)


injms:
Do not merge label added, as I haven't run it through any accessibility or browser testing yet.

injms:
All testing done; ready for review.

app/views/download/help_improve.html.haml
```diff
@@ -85,8 +81,164 @@
...
+                readFormFn.sendData(form);
+                form.submit();
+              }
+              break
+            case false:
+            default:
+              errorFn.showErrorWarning(form, {
+                addClassTo: '.js--is-government-group',
+                putMessageIn: '.js--is-government-message',
+                messageShouldRead: 'Select yes if you work for the government'
```
gidsg: I think this should be _Select if you work for the government_ rather than leaning towards one answer.

injms: I've moved the error messages to the i18n config file. 

The wording follows the guidelines in the design system and Mark has okayed it.

PR #484, opened 2018-11-06
Adds secondary navigation

injms:
### Context
We currently only have the header and footer to allow users to navigate through frontend.

As content grows we need to have an internal navigation functionality that allows users access to priority content that is not covered in the header.

* Secondary navigation partial has been created and included on the homepage, about page, and registers collection page
* The 'registers in numbers' has been modified and is included on the same pages.
* The numbers appear more prominently on the homepage, but smaller on other pages.

### Guidance to review
None.


app/views/layouts/_registers-in-numbers.html.haml
```diff
@@ -1,19 +1,21 @@
...
-    %li{class: 'registers-in-numbers__item govuk-!-margin-bottom-4'}
-      -# = link_to registers_path do
-      %span{role: "presentation", class: "#{current_page?(about_path) ? 'govuk-heading-xl' : 'govuk-!-font-size-80 govuk-!-font-weight-bold'} govuk-!-margin-bottom-0"}= Register.organisation_count
-      %span{role: "presentation", class: "#{current_page?(about_path) ? 'govuk-body-s' : 'govuk-body'} govuk-!-margin-bottom-0 data-item"} government organisations managing&nbsp;registers
-    %li{class: 'registers-in-numbers__item govuk-!-margin-bottom-4'}
-      -# = link_to services_using_registers_path do
-      %span{role: "presentation", class: "#{current_page?(about_path) ? 'govuk-heading-xl' : 'govuk-!-font-size-80 govuk-!-font-weight-bold'} govuk-!-margin-bottom-0"} 8
-      %span{role: "presentation", class: "#{current_page?(about_path) ? 'govuk-body-s' : 'govuk-body'} govuk-!-margin-bottom-0 data-item"} government services using&nbsp;registers
+- list_presentation_class = defined?(presentation).nil? ? '' : " registers-in-numbers--#{presentation}"
+- heading_border_class = defined?(border).nil? ? '' : " registers-!-border-top"
```
MatMoore: I don't think your nil check is working here because it's called on the result of `defined?`, which is a boolean. I would normally avoid the use of `defined?` checks entirely in ruby - it makes the code more complicated if your variable can either be set, nil, or undefined.

Instead I would make sure the variable already is defined by the time you are rendering the template, and then you can use `present?` or `absent?` in the template. Would it make sense to always pass through the class names from the parent template?

I also found this stack overflow which suggests some other ways to provide optional values to a partial, not sure if it's helpful or not. https://stackoverflow.com/questions/2060561/optional-local-variables-in-rails-partial-templates-how-do-i-get-out-of-the-de

app/views/layouts/_registers-in-numbers.html.haml
```diff
@@ -1,19 +1,21 @@
...
+- heading_size = defined?(heading).nil? ? 'govuk-heading-m' : 'govuk-heading-s'
+
+%h2{ class: "#{heading_size} #{heading_border_class} govuk-!-padding-top-4"} GOV.UK Registers in numbers
+%ul{class: "registers-in-numbers#{list_presentation_class}" }
+  %li{class: 'registers-in-numbers__item'}
+    %span{role: 'presentation', class: 'registers-in-numbers__number'} 800k
+    %a{role: 'presentation', class: 'registers-in-numbers__caption', href: 'https://www.gov.uk/performance/govuk-registers'}total API requests
+  %li{class: 'registers-in-numbers__item'}
+    -# = link_to registers_path do
```
MatMoore: If this is commented out can we remove it now?

app/views/layouts/_registers-in-numbers.html.haml
MatMoore: Btw, what does the `!` in the css class signify? I'm not familiar with the naming conventions.

app/views/layouts/_registers-in-numbers.html.haml
injms: Removed 👍

app/views/layouts/_registers-in-numbers.html.haml
injms: I wanted a default layout, with the variables only being defined to add CSS classes the expand upon those initial rules. I've move the variables to the controller and changed the logic that checks whether they're present - is the new version okay?

The `!` shows that `!important` is being used to boost the specificity of the rule in that class - the [design system](https://design-system.service.gov.uk/styles/typography/#font-override-classes) has a few of these override classes, and I thought it quite elegant.

PR #485, opened 2018-11-07
Removes 'registers' from each event being fired, making them less wordy.

injms:
### Context
Cliff requested that the events that are when going from a collection / organisation / list page to the register page are changed.

* 'register' to be removed from the end of the event actions; for example, _Organisation - Ministry of Justice - Prison estate register_ to be changed to _Organisation - Ministry of Justice - Prison estate_

### Guidance to review
None.

### DIFF
--- a/app/views/authority/show.html.haml
+++ b/app/views/authority/show.html.haml
@@ -28,7 +28,7 @@
         ga('send', {
           hitType: 'event',
           eventCategory: 'Show by',
-          eventAction: 'Organisation - #{@collection.name.html_safe} - ' + event.target.innerText,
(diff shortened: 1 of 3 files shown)


app/views/authority/show.html.haml
```diff
@@ -28,7 +28,7 @@
         ga('send', {
           hitType: 'event',
           eventCategory: 'Show by',
-          eventAction: 'Organisation - #{@collection.name.html_safe} - ' + event.target.innerText,
+          eventAction: 'Organisation - #{@collection.name.html_safe} - ' + event.target.innerText.replace(/ register/, ''),
```
gidsg: can we use end of string regex i.e. `^` to make sure this does not erroneously do the wrong replace?

injms:
Yes, this change was requested by Cliff.

PR #486, opened 2018-11-07
Changes to register record history; makes history more prominent

injms:
### Context
We want to evaluate public demand for the audibility of registers. The provision of this functionality is currently available via the 'Last updated link' on individual register page. This functionality need to be made more prominent.

* Link added below accordion to register history
* Events fired when the metadata (organisation, category, last updated) links are clicked on
* History page edits - register name added, space between last updated and date; lowercase 'v' on values.

### Guidance to review
None.

### DIFF
--- a/app/assets/stylesheets/modules/_subsections.scss
+++ b/app/assets/stylesheets/modules/_subsections.scss
@@ -163,7 +163,6 @@
 }
 
 .register-show.js-enabled main {
-  // border-bottom: 1px solid $border-colour;
   padding-bottom: 0 !important;
   margin-bottom: 90px !important;
 }


app/assets/stylesheets/modules/_subsections.scss
```diff
@@ -163,7 +163,7 @@
 }
 
 .register-show.js-enabled main {
-  border-bottom: 1px solid $border-colour;
+  // border-bottom: 1px solid $border-colour;
```
MatMoore: Can we remove this line?

injms: Done.

PR #487, opened 2018-11-08
Corrects registers use reporting

injms:
### Context
GA was reporting the government department instead of the register use - this patch corrects that.

* Selector changed to `gov_what_are_you_using_registers_for`


app/views/download/help_improve.html.haml
```diff
@@ -46,7 +46,7 @@
         sendData: function(form) {
           var isGovernment = readFormFn.checkAnswer(form, 'help_us_improve_user[is_government]')
           var whatPartOfGovernmentAreYouWorkingFor = readFormFn.checkAnswer(form, 'help_us_improve_user[gov_what_part_of_government]')
```
gidsg: would it be more understandable if we named these variables closer to our model rather than the form wording? e.g. `department` rather than `whatPartOfGovernmentAreYouWorkingFor`

injms: Agreed - however, this patch needs to go out to fix the reporting as soon as possible. I'll raise another pull request to refactor the naming convention.

PR #496, opened 2018-11-15
Offer ODS downloads 

gidsg:
### Context
We need to offer ODS downloads as we are deprecating XLSX

Serialize to ODS from frontend DB

### Guidance to review
You should be able to download a register as ODS following the frontend download journey.
Still to do on this is to implement Analytics, need spec from @cliffshepp 
@mpwoods should also probably look over wording.


app/services/register_records_downloader.rb
```diff
@@ -19,6 +19,14 @@ def download_format(format)
     end
   end
 
+  def download_ods
+    headers = register.fields_array
+    data = register.records.where(entry_type: 'user').find_each.map do |r|
```
MatMoore: you could use activerecord's `pluck` here to avoid instantiating active record objects for each record. The trade off is you can't then batch things using `find_each`, but in this case you're reading the whole dataset into memory anyway so I don't think it matters.

app/views/download/index.html.haml
```diff
@@ -12,5 +12,5 @@
       %p You can download the latest data in this register as:
       %ul.list.list-bullet{data: {"click-events" => true, "click-category" => "Content", "click-action" => "Register Download"}}
         %li= link_to 'CSV', register_download_csv_path(@register.slug)
-        %li= link_to 'JSON', register_download_json_path(@register.slug)
+        %li= link_to 'ODS (Excel compatible)', register_download_ods_path(@register.slug)
```
MatMoore: 👍 I think it's good idea to take JSON out here

spec/controllers/download_controller_spec.rb
```diff
@@ -0,0 +1,44 @@
+require 'rails_helper'
+
+RSpec.describe DownloadController, type: :controller do
```
MatMoore: Can we avoid stubbing out RSF and running the importer as part of the setup, and just create some database fixtures using FactoryBot? Or if not, maybe extract this setup to a common test helper (I think there are a few others that follow the same pattern)?

In this test we're not actually testing the contents of the spreadsheet, so a lot of the code in this file is incidental what we're testing.

On that note, maybe its worth creating a unit test that parses the resulting ODS and checks the data and headings are correct? I'm not too worried about this though.

app/services/register_records_downloader.rb
```diff
@@ -19,6 +19,14 @@ def download_format(format)
...
+      headers.map { |f| r.data[f] }
+    end
+    SpreadsheetArchitect.to_ods(headers: headers, data: data)
```
MatMoore: Is there anything else we want to provide in this spreadsheet? For example, a different sheet with metadata in it?

arnau: Not at the moment. The goal of this format is to provide an alternative to CSV that can be open without ceremony in UTF-8.

spec/controllers/download_controller_spec.rb
gidsg: @MatMoore Yes, I agree ideally we would parse the contents of the ODS file unfortunately this would require another test dependency that can parse this format so I'm not sure it's worth the effort.

I will look at moving the test boilerplate out somewhere.

gidsg:
@MatMoore 
re https://github.com/openregister/registers-frontend/pull/496#pullrequestreview-175365018
1) Yes, I agree in principal we should move CSV to this mechanism. In practice I would rather release ODS first and do a separate PR for CSV, this is so we have some production traffic going into GA to prove the new mechanism works equivalently for ODS before swapping CSV over.

2) This mechanism is fast enough for the largest register we have today. I agree we could for example write them to S3 as part of the indexing process but I think this method is good enough for now and we can re-visit if we have a very large register in the pipeline.

MatMoore:
Sounds good to me.

Can we create a backlog card to look at the CSV? I think it could cause confusion when it comes to measuring CSV downloads vs ODS downloads because at the moment I think we report on downloads based on the ORJ requests (i.e. it's tracked as a kind of API call rather than as a click in frontend).

spec/controllers/download_controller_spec.rb
gidsg: OK, I moved out the RSF importing now in: https://github.com/openregister/registers-frontend/pull/496/commits/d1c02054b6e41300d6b81f12396254f0f291b597

gidsg:
@MatMoore I have added a planning card for generating CSV in the frontend.

gidsg:
OK, I am going to merge this now, will make any wording changes suggested by @mpwoods in a subsequent PR.

PR #506, opened 2018-11-26
New about page

injms:
### Context
The _About_ page content needs to be updated to create a quick guide to GOV.UK Registers.

See the [card](https://trello.com/c/osHuY84g/2978-quick-guide-to-registers) for more details.

* Updates _About_ page with content supplied by @mpwoods and @stephenjoe1 
* Adds styles to create an image block that doesn't jank
* Removes `application.scss` file - empty file that wasn't used.

### Guidance to review
* Content has been okayed by Mark.


app/views/pages/about.html.haml
```diff
@@ -11,22 +11,41 @@
...
+        %p You can see all the registers that are ready to use in our #{link_to 'government registers collection', registers_path}.
+
+        %h3.govuk-heading-m Who is using registers
+        %p You can see which #{link_to 'government services', services_using_registers_path} are already using registers.
+        %p You can also find out which #{link_to 'government organisations', registers_path(show_by: "organisation")} have created a register with their data.
+
       %nav{class: 'govuk-grid-column-one-third'}
         = render partial: 'layouts/secondary-navigation'
-      %aside{class: 'govuk-grid-column-two-thirds govuk-!-padding-top-4'}
-        = render partial: 'layouts/registers-in-numbers'
```
gidsg: Is this partial still used somewhere else or can it be deleted?

injms: Good spot - thank you. It's now deleted.

gidsg: Seems like there are some other references e.g. https://github.com/openregister/registers-frontend/blob/5950aabbda8f87aacc220b2205b6a94d0e7fb5ff/app/controllers/pages_controller.rb#L29 please can you check if they can be safely deleted?

injms: The only place those variables were used was in the now-removed partial - so I've removed the variables as well.

injms: Um - The partial was only being used in one place 😳 

I've removed the file and placed the code in the template, plus refactored the variables that no longer need to be passed from the controller. It'll be easy enough to revert this if needed in the future.

PR #511, opened 2018-11-27
Analytics distinguishes between live and alpha registers

injms:
### Context
Cliff needs to be able to distinguish events and page views based on whether the register is live or in alpha.

* [Card](https://trello.com/c/IqsrInDU/3037-make-distinction-between-registers-in-development-and-live-registers-for-analytics).

* Cleans up existing custom dimension naming
* Adds custom dimension only on pages associated with a register (for example, the history, questionnaire, and download pages)
* Custom dimension added to analytics partial, so every event and page view will have that custom dimension attached

### Guidance to review
None.


app/controllers/download_controller.rb
```diff
@@ -10,7 +10,8 @@ class DownloadController < ApplicationController
   helper_method :government_orgs_local_authorities
 
   def index
-    @custom_dimension = "#{@register.name} - download"
+    @custom_dimension2 = "#{@register.name} - download"
+    @custom_dimension3 = @register.register_phase == 'Alpha' ? 'Alpha' : 'Live'
```
gidsg: This will erroneously label `Discovery`  registers as `Live`.  If we really have to change `Beta` to `Live` can we have a mapping that incorporates all phases? Or just use the phase as-given if @cliffshepp can handle that?

injms: From discussion, I was under the impression that `Discovery` registers shouldn't be present on the frontend app - hence the simpler ternary operator here. Is that an incorrect assumption?

gidsg: Yes, there are a bunch of `Discovery` registers in the production CMS so I don't think we can rely on them not being present.

app/controllers/download_controller.rb
```diff
@@ -26,6 +27,8 @@ def create
   def success; end
 
   def choose_access
+    @custom_dimension3 = @register.register_phase == 'Alpha' ? 'Alpha' : 'Live'
```
gidsg: we are repeating this logic multiple times, could we DRY it by moving it to a method on the model or something?

app/controllers/download_controller.rb
injms: Okay - I'll move the logic into its own method and use that to label the custom dimension with one of the three phases.

app/controllers/download_controller.rb
injms: Done 👍

PR #513, opened 2018-11-28
Changes to register page header and meta

injms:
### Context
First in the set of changes to the register page, as [detailed in the card](https://trello.com/c/Kn6y0mKh/3094-implement-new-access-the-data-options-individual-register-page-changes). 

This changes the layouts of the register page header and refines what metadata is shown.

 * Publishing organisation removed from metadata
 * First published date added to metadata; added method for getting the date
 * CSS for previous metadata layout removed
 * Date format changed, including helper method to make it human readable

### Guidance to review
 * This is only part one of the changes asked for - this is to avoid a mega-deploy.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -40,7 +40,11 @@ module ApplicationHelper
   end
 
   def formatted_date(date)
-    Date.parse(date).strftime('%d/%m/%Y')
+    Date.parse(date).strftime('%d/%m/%Y') # eg 03/08/2015
+  end
+
+  def human_readable_date(date)
+    Date.parse(date).strftime('%e %B %Y') # eg 3 August 2015
   end
 
   def records_table_header(field_value)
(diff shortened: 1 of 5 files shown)


app/helpers/application_helper.rb
```diff
@@ -40,7 +40,11 @@ def phase_label(phase)
   end
 
   def formatted_date(date)
-    Date.parse(date).strftime('%d/%m/%Y')
+    Date.parse(date).strftime('%d/%m/%Y') # eg 03/08/2015
+  end
+
+  def human_readable_date(date)
+    Date.parse(date).strftime('%e %B %Y') # eg 3 August 2015
```
MatMoore: Another way of doing this is to use a relative date like "over 3 years ago". I think this is the rails method for doing that: https://apidock.com/rails/ActionView/Helpers/DateHelper/time_ago_in_words

I don't know which is more useful in our case though.

injms: The '3 August 2015' format was what was requested - but that `time_ago_in_words` is very neat indeed! A lot of very nice helpers in Ruby.

PR #514, opened 2018-11-30
Removes feedback form from register page

injms:
### Context
The feedback form was setting extremely limited usage and any comments coming through would better served by the existing contact form. This is part of the [access to data](https://trello.com/c/DfDyseKq/112-access-the-data-options-prominent-on-individual-register-page) changes; here's the [ specific card](https://trello.com/c/F4Z1gMNj/3131-remove-feedback-form-on-individual-register-pages) for the removal of the feedback form.

* Deletes form partial
* Deletes feedback model
* Removes inclusion of form partial from register page
* Removes methods and variables from the register controller that were used in the feedback form

### Guidance to review
* Please check that I haven't left anything that'll be unused
* I've checked that this doesn't break any other pages, but a second opinion would be very reassuring


gidsg:
@injms  Seems from this earlier PR there might be some CSS that is safe to delete now, worth checking at least:
https://github.com/openregister/registers-frontend/pull/323/files#diff-f859e2848b07324f808ce1a5ccd9fcba

injms:
> @injms Seems from this earlier PR there might be some CSS that is safe to delete now, worth checking at least:
> https://github.com/openregister/registers-frontend/pull/323/files#diff-f859e2848b07324f808ce1a5ccd9fcba

Thanks for pointing me in this direction - I also found the route and i18n file could be cleaned up as well. Cheers!

PR #518, opened 2018-12-03
Makes register download / API options more prominent

injms:
### Context
Getting access to a Register's data means a user needs to choose whether they want to download the data or use the API. This choice was the second step in the flow - but surfacing the choice sooner highlights the different methods offered and removes a step in the process. 

The [card](https://trello.com/c/OO0dvpJl/3130-implement-new-access-the-data-options-on-individual-register-page) has the prototype and the content documents attached to it.

* Removes the 'Choose access' step
* Adds API and download options onto register page
* Removes custom form header styles
* Replaces old highlight boxes styles with new
* Small amount of refactoring, removing old Elements and adding new Design System
* Rewrote tests to take into account new flow

### Guidance to review
* Look out for anything else that could be removed
* Layout uses `display: flex`, so falls back to floats for Internet Explorer 8 to 10


app/views/download/_form--help-us-improve.html.haml
```diff
@@ -50,7 +50,7 @@
   %div{ class: "govuk-form-group access-to-data__submit-group"}
     %button{type: 'submit', class: 'govuk-button govuk-!-margin-bottom-0'} Submit
     %span or
-    = link_to "I'd prefer not to say", @next_page, { class: 'js--ga-watch-click' }
+    = link_to "Skip this step", @next_page, { class: 'js--ga-watch-click' }
```
gidsg: very minor, we can use single quotes here as we're not doing string interpolation.

injms: Fixed 🔍

PR #519, opened 2018-12-03
filter category listing to registers with records only

gidsg:
### Context
Previously when a register was added to the CMS but had not yet been indexed it would be linked to from the category list and cause a 404

only show registers with records in category list

### Guidance to review
Add a beta register with a category in the CMS
It should not appear in the CMS until the populate data script has been run

### DIFF
--- a/app/models/authority.rb
+++ b/app/models/authority.rb
@@ -12,7 +12,7 @@ class Authority < ApplicationRecord
   }
 
   def registers_by_this_collection
-    @registers_by_this_collection ||= registers.in_beta.by_name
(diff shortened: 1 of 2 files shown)


injms:
Just realised that the sorting by name had gone, and the count displayed for each category / authority was using including empty beta registers. The last commit should fix these things.

gidsg:
@injms probably the stuff in https://github.com/openregister/registers-frontend/pull/519/commits/3b9703a3a366f657c7d4aacaa895fe3a92ee568b will come when conflicts are fixed.

PR #522, opened 2018-12-07
New tables for registers page

injms:
### Context
The page for each register contains a table showing the data contained in that register. This is being updated to only show a set number of records. A further table is added to show the register's fields, plus an example of each field from that register.

This should follow the layout in the prototype, as linked to from the [card](https://trello.com/c/p0jeQjUc/3133-implement-new-field-descriptions-and-preview-on-individual-register-pages).

* table of records limited to 5 on all registers, except those published by DCMS which is limited to 10
* 'load more' or pagination only present on DCMS published registers
* Question mark (What is...?) removed from table header
* Table showing fields and description for each field added
* Removes search; removes search tests

### Guidance to review
* No DCMS registers are on the staging site or on the local seeds; give me a shout and we can copy the live database to your local install


app/models/register.rb
```diff
@@ -79,6 +79,37 @@ def show_category_link?
...
+    end
+  end
+
+  def safe_name
+    # `seo_title` already has 'register' in it. So we need to replace 'register
+    # register' with a 'register'.
+    [[title, seo_title, name].find(&:present?), 'register']
+      .compact
+      .join(' ')
+      .gsub('register register', 'register')
```
gidsg: We should probably check this is the end of the string, as this would do the replacement anywhere currently.

app/views/layouts/_list-of-registers.html.haml
```diff
@@ -2,6 +2,6 @@
   - @collection.registers_by_this_collection.each do |register|
     %li.list-of-registers__register
       %h3.list-of-registers__heading
-        =link_to "#{register.title} register", register_path(register.slug), {class: 'govuk-heading-s govuk-!-margin-bottom-0'}
+        =link_to "#{register.safe_name}", register_path(register.slug), {class: 'govuk-heading-s govuk-!-margin-bottom-0'}
```
gidsg: do we have a decision on when to use the SEO title and when to use the register title? As currently, this will put the SEO title in more places.

app/views/registers/_fields.html.haml
```diff
@@ -0,0 +1,23 @@
+
+%table.registers-secondary-table
+  %caption.registers-secondary-table__caption
+    The
+    <!-- -->
+    %b #{@register.register_fields.count} fields
+    <!-- -->
```
gidsg: Can we move these comment blocks?

app/views/registers/show.html.haml
```diff
@@ -37,25 +36,25 @@
...
-          = search_field_tag 'q', nil, class: 'search-input', placeholder: 'Search', value: @search_term, autocomplete: 'off'
-          = submit_tag 'Search', name: nil, class: 'search-submit'
-      .govuk-grid-column-full.subsection__content.js-accordion-subsection#about-the-data__2
+      .govuk-grid-column-full
+        %hr{class: 'govuk-!-margin-top-0'}
+        %h2.govuk-heading-m About the data
+      %div{class: 'govuk-grid-column-two-thirds govuk-!-margin-bottom-3'}
+        %p
+          There are
+          <!-- -->
```
gidsg: can we remove these comment blocks?

app/views/registers/show.html.haml
```diff
@@ -71,62 +70,35 @@
...
-            %script
-              ga('send', 'event', 'Search', 'No results', '#{@search_term}', {nonInteraction: true});
-          - else
-            %script
-              ga('send', 'event', 'Search', 'No results', 'Search term not present', {nonInteraction: true});
-    %hr{class: 'govuk-!-margin-top-0'}
-%dialog#modal_dialog.modal-dialog.dialog{role: "dialog", "aria-hidden" => "true", "aria-labelledby" => "dialog-title"}
-  %button{type: 'button', class: 'modal-dialog__close', "aria-label" => "Close dialog"} Close
-  .modal-dialog__inner
+          - if @show_load_more == true
```
gidsg: I *think* the `== true` is redundant here as `if @show_load_more` would resolve to the same value.

config/locales/en.yml
```diff
@@ -32,7 +32,7 @@ en:
           one: "Showing <b>1</b> %{entry_name}"
           other: "Showing <b>%{count}</b> %{entry_name}"
       more_pages:
-        display_entries: "Showing <b>%{first}&ndash;%{last}</b> of <b>%{total}</b> %{entry_name}"
+        display_entries: "<b>Preview of records %{first} &ndash; %{last}</b>"
```
gidsg: I am not sure that this text is strictly accurate since.:
* Records has no ordering
* We provide a default display order of either `key` or `name`.

Perhaps we should just say *showing* as before, or maybe just 
```<b>Preview of records %{last} records</b>```
?

CC @mpwoods

Comment (mpwoods) [reply config/locales/en.yml]: Let's go with 'Preview of 5 records'

File: app/models/register.rb [hunk 7]
```diff
@@ -79,6 +79,37 @@ def show_category_link?
     register_phase == 'Beta' && category.present?
   end
 
+  def is_register_published_by_dcms?
+    Authority.where(id: authority_id).first&.government_organisation_key === 'D5'
+  end
+
+  def fields_readable
+    register_fields.map do |field|
+      # Make the field more readable.
```
gidsg: rather than assigning a new property to fields I would rather we have a function that applies the transformation to a single field, then call that where needed.

app/models/register.rb
```diff
@@ -79,6 +79,37 @@ def show_category_link?
...
+      field['field_readable'] = field['field']
+                                  .humanize
+                                  .tr('-', ' ')
+
+      # GOV UK URLs are the only thing that seems to break the table on small
+      # screens. This inserts zero-width breaking spaces (`&#8203;`) to allow
+      # a URL to wrap in a narrow table cell.
+      field['text_readable'] = field['text']
+                                .gsub('.service.', '&#8203;.service.&#8203;')
+                                .html_safe
```
gidsg: is there any way we can do this more generically, as there are other registers which might have long URLs?

app/views/layouts/_list-of-registers.html.haml
injms: This method was added because there were places on staging and on local that `title` wasn't being populated and `seo_title` gave a better fallback. 

I'd like to raise a ticket to streamline the names once and for all, but in the meantime I think that this method provides an acceptable stopgap.

app/views/registers/_fields.html.haml
injms: Unfortunately not if we want spaces between the words - Haml is set up to strip out whitespace, even here when you've got nested elements.

app/views/registers/show.html.haml
injms: Same as the other instance of this - the comment blocks preserve a space between the words.

config/locales/en.yml
injms: Changed; I've also changed it to this phrasing for the register that are able to load more records . Where there are less than the limit it'll say 'Showing 3 records', since that isn't really a preview.

app/controllers/registers_controller.rb
```diff
@@ -35,11 +35,15 @@ def in_progress
   end
 
   def show
-    @search_term = search_term
     @register = Register.has_records.find_by_slug!(params[:id])
+
     @records = recover_records(@register.fields_array, params)
+    @records_unfiltered = recover_records(@register.fields_array, params, true)
```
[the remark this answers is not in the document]
injms: Nope, not now we've got rid of the examples.

app/models/register.rb
injms: Fixed.

app/models/register.rb
injms: I've spun these into their own functions in the `application_helper.rb` file so they can be called when needed.

app/models/register.rb
injms: I've given it a shot and put it in `application_helper` - had to make a couple of reasonable (I think) assumptions that it'd be good to get your thoughts on.

app/views/layouts/_list-of-registers.html.haml
injms: [Ticket raised]( https://trello.com/c/X1LeLGTf/1559-remove-the-uncertainty-around-title-seotitle-and-name) 👍

app/views/registers/_fields.html.haml
gidsg: can we just use in-line HTML instead e.g. 
```
The <strong>#{@register.register_fields.count}</strong> fields in this register
```

app/helpers/application_helper.rb
```diff
@@ -47,13 +47,30 @@ def human_readable_date(date)
...
+    # Assumes a full stop with a lowercase letter, number, or an underscore on
+    # both sides is part of a URL; then adds an invisible breaking space to
+    # allow wrapping.
+    # eg:
+    #  - Apples. Oranges. Pears. => Apples. Oranges. Pears.
+    #  - GOV.UK => GOV.UK
+    #  - registers.service.gov.uk => registers.&#8203;service.&#8203;gov.&#8203;uk
+    string
+      .gsub(/([a-z0-9_])\.([a-z0-9_])/, '\1.&#8203;\2')
+      .html_safe
```
gidsg: What's the thinking behind this approach rather than using CSS word-wrap?

injms: `overflow-wrap: break-word` is fine for a long word that goes outside of the table element - but I think that breaking on the `.` of the URL helps keep the table more readable. For example, with `break-word` you could end up with: 
```
registers.se
rvice.gov.uk
```
By adding an invisible space, you get:

```
registers.
service.
gov.uk
```

app/views/registers/_fields.html.haml
injms: Sure - using comments was something that was used elsewhere in this repo, so I was trying to keep thing consistent. If I see the other uses of this I'll change them to inline as it is more readable.

app/models/register.rb
```diff
@@ -79,6 +79,19 @@ def show_category_link?
...
+    Authority.where(id: authority_id).first&.government_organisation_key === 'D5'
+  end
+
+  def safe_name
+    # `seo_title` already has 'register' in it. So we need to replace 'register
+    # register' with a 'register'.
+    [[title, seo_title, name].find(&:present?), 'register']
+      .compact
+      .join(' ')
+      .gsub(/register register$/i, 'register')
```
gidsg: I am not sure all the contexts where `seo_title` is used but maybe we can just remove the hardcoded `register` bit from them rather than hacking it in the code. This is OK for now though.

injms: Happy to clean up the title / naming spaghetti later - I've raised a [ticket for that](https://trello.com/c/X1LeLGTf/1559-remove-the-uncertainty-around-title-seotitle-and-name).

app/models/register.rb
```diff
@@ -79,6 +79,19 @@ def show_category_link?
     register_phase == 'Beta' && category.present?
   end
 
+  def is_register_published_by_dcms?
+    Authority.where(id: authority_id).first&.government_organisation_key === 'D5'
```
gidsg: I think we can express this more simply as: 
```
authority.government_organisation_key == 'D5'
```

gidsg: I think it might work if you use safe navigation operator e.g.
```
authority&.government_organisation_key == 'D5'
```

injms: Cheers!

PR #529, opened 2018-12-17
Adds tests to check the access to data flow works correctly.

injms:
### Context
The Access to data flow should have tests - this is to make sure both download and API flows are working, and that the questionnaire is shown only on the first visit through the flow. This will make sure that this flow is working without requiring unreliable manual testing.

[Card](https://trello.com/c/QnOSyqSS/3158-create-feature-test-for-access-to-data-flow).

  * Checks both API and download paths are working
  * Checks questionnaire
  * Checks questionnaire is skipped if it's already been seen
  * Checks content on flow

### Guidance to review
* The new tests should run with the command `bundle exec rake spec`
* The balance between making a test easy to understand and keeping it all DRY - this falls on the side of not being DRY and (hopefully) being easier to understand. Improvements on that welcome
* Let me know if I've missed anything that should be tested

### DIFF
--- a/spec/features/access_data_flow_spec.rb
+++ b/spec/features/access_data_flow_spec.rb
@@ -47,10 +47,8 @@ RSpec.feature 'View register', type: :feature do
   scenario 'the states for all pages in API flow are 200' do
     visit('/registers/country')
     expect(page.status_code).to eq(200)
-
     click_link('Use the API')
     expect(page.status_code).to eq(200)
-
     click_link('Skip this step')
     expect(page.status_code).to eq(200)
   end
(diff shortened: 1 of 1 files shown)


spec/features/access_data_flow_spec.rb
```diff
@@ -39,25 +39,109 @@
...
+  end
+
   scenario 'goes to questionnaire correctly when download option chosen' do
     visit('/registers/country')
     click_link('Download the data')
 
     expect(page).to have_content('Before you download the data')
     expect(page).to have_content('Help us improve GOV.UK Registers')
+
+    expect(page).to have_content('What part of government are you working for?')
```
gidsg: I think we should be able to use `include` and make an array with all the expected values: https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/include-matcher

PR #536, opened 2018-12-31
Adds Markdown rendering for records

injms:
### Context
The frontend was rendering the text datatype as plain text rather than markup. This led to any Markdown not being rendered - see the [datatype register](https://www.registers.service.gov.uk/registers/datatype) as a good example of this. The expected behaviour is that any text datatype is rendered as Markdown.

Barebones [card](https://trello.com/c/1e8Q5d8r/3175-we-do-not-render-markdown-in-fields-with-text-datatype).

* Adds Bluecloth Markdown renderer to the gem
* Edits the `field_link_resolver` in the registers controller to add Markdown parsing
* Updates tests

### Guidance to review
* Remember to `bundle install` to get Bluecloth
* Registers with interesting text fields that are worth checking:
  * Datatype register
  * Register register
  * Approved open standard guidance register

### DIFF
--- a/app/controllers/registers_controller.rb
+++ b/app/controllers/registers_controller.rb
@@ -2,7 +2,7 @@
 
 class RegistersController < ApplicationController
   include ActionView::Helpers::UrlHelper
-  helper_method :field_link_resolver
+  helper_method :field_formatter
 
 
   def index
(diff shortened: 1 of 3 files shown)


app/controllers/registers_controller.rb
```diff
@@ -56,9 +57,11 @@ def field_link_resolver(field, field_value, register_slug: @register.slug, white
     resolver = LinkResolver.new(current_register_slug: register_slug, register_whitelist: whitelist)
 
     if field_value.is_a?(Array)
-      field_value.map { |fv| resolver.resolve(field, fv) }.join(', ').html_safe
+      list = field_value.map { |fv| resolver.resolve(field, fv) }.join(', ')
+      BlueCloth.new(list, auto_links: false).to_html
```
gidsg: we should only apply this where we know the field may contain markdown. So we can do something like:
```
    with_links = if field_value.is_a?(Array)
                   field_value.map { |fv| resolver.resolve(field, fv) }.join(', ').html_safe
                 else
                   resolver.resolve(field, field_value)
                 end
    field['datatype'] == 'text' ? BlueCloth.new(with_links, auto_links: false).to_html : with_links
```

gidsg:
Can we rename `field_link_resolver` to `field_formatter` or something, now it is doing multiple things?

app/controllers/registers_controller.rb
injms: Done. The linter was complaining about misalignment of `if`, `else`, and `end` so I've used a ternary instead.

gidsg: OK cool, feel free to merge. By the way I don't believe you need `===` in this case for the comparison.

PR #576, opened 2019-04-16
use full reload task for scheduled job

gidsg:
### Context
https://trello.com/c/t3h6bwAq/3234-frontend-rsf-consumption-rework

Use Full Reload task for scheduled job. Remove incremental download task.

### Guidance to review
I have tested this locally using the queue and confirmed using a proxy it only makes requests to `/download-rsf/0`. 

You can check `rake` command in README still works locally to populate data


phantas:
LGTM but before I approve,  have you tested this with our biggest register? Any issues holding it all in memory, or runtime of the job?

gidsg:
@phantas we do not have the queue/scheduler running in sandbox but I have tested this works successfully using `cf run-task` with the same amount of memory which I believe to be a reasonable proxy e.g. 
```
cf run-task registers-frontend-research "bundle exec rake registers_frontend:populate_db:fetch_now" -m 1G
```

PR #653, opened 2020-01-03
Bau/switch off non essential cookies

phantas:
### Context

Due to recent cookie policy changes, we are switching off any non essential cookies. Those are:
- GA cookies
- seen_cookie_message

- Update Cookie policy to remove mentions of non-essential cookies
- Don't set Google Analytics cookies and don't send data to GA
- Hijack application rendering to remove cookie banner (also removes seen_cookie_message)

See commit messages for more details

### Guidance to review

- See commit messages
- checkout branch locally and run server (don't use Docker, which is currently broken)
- Open an Incognito Message
- Open localhost:3000
- Confirm you don't see the cookie banner
- Confirm only the session cookie is set
- Open /cookies and confirm policy no longer mentions non-essential cookies
- Navigate to a register, fill in questionnaire
- Confirm you see the corresponding cookie and are able to download a register without issues


app/controllers/application_controller.rb
```diff
@@ -11,5 +11,12 @@ class ApplicationController < ActionController::Base
   def current_user
     @current_user ||= User.where(id: session[:user_id]).first if session[:user_id]
   end
-  helper_method :current_user
+
+  def remove_cookie_banner(s)
```
freesteph: I would call this parameter something a bit more explicit like `markup`, or `text`.

app/controllers/application_controller.rb
```diff
@@ -11,5 +11,12 @@ class ApplicationController < ActionController::Base
...
+    #  Hot patch to remove the cookie banner before rendering
+    cookie_banner_regex = %r{<div id="global-cookie-message">\s*My Cookie Message\s*<\/div>}
```
freesteph: Can we leverage the fact this is a regex and use a wildcard instead of `My Cookie Message`? This means that should this ever fail (for reasons we're sadly unaware of) the default copy is still displayed and is probably more informative for our users.

app/services/register_records_downloader.rb
```diff
@@ -20,7 +20,7 @@ def download_format(format)
   end
 
   def download_ods
-    ods_download_to_ga
+    #ods_download_to_ga
```
freesteph: for all of the GA stuff, does it make sense to delete the code instead? This way we can restore them with a quick git-revert, and no dead code lying around.

app/controllers/application_controller.rb
phantas: The wildcard wasn't working for me yesterday. Figured out today that there were whitespaces preventing it, so changed back

freesteph: An alternate solution: 

```diff
diff --git a/app/controllers/application_controller.rb b/app/controllers/application_controller.rb
index 85e2e38..c8cb036 100644
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -13,9 +13,10 @@ private
   end
 
   def remove_cookie_banner(content)
-    #  Hot patch to remove the cookie banner before rendering
-    cookie_banner_regex = %r{<div id="global-cookie-message">\s*My Cookie Message\s*<\/div>}
-    content.gsub(cookie_banner_regex, '')
+    doc = Nokogiri::HTML.fragment(content)
+    doc.search('#global-cookie-message').remove
+
+    doc.to_html
   end
 
   helper_method :current_user, :remove_cookie_banner
```

app/controllers/application_controller.rb
phantas: I caught it on linting. Renamed to content

app/controllers/application_controller.rb
```diff
@@ -11,5 +11,12 @@ class ApplicationController < ActionController::Base
   def current_user
     @current_user ||= User.where(id: session[:user_id]).first if session[:user_id]
   end
-  helper_method :current_user
+
+  def remove_cookie_banner(content)
+    #  Hot patch to remove the cookie banner before rendering
+    cookie_banner_regex = %r{<div id="global-cookie-message">\s*.*\s*<\/div>}
+    content.gsub(cookie_banner_regex, '')
+  end
```
freesteph: Alternate solution, for reference:

```diff
diff --git a/app/controllers/application_controller.rb b/app/controllers/application_controller.rb
index bf89645..c8cb036 100644
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -13,9 +13,10 @@ private
   end
 
   def remove_cookie_banner(content)
-    #  Hot patch to remove the cookie banner before rendering
-    cookie_banner_regex = %r{<div id="global-cookie-message">\s*.*\s*<\/div>}
-    content.gsub(cookie_banner_regex, '')
+    doc = Nokogiri::HTML.fragment(content)
+    doc.search('#global-cookie-message').remove
+
+    doc.to_html
   end
 
   helper_method :current_user, :remove_cookie_banner
```

phantas:
> Happy with it 👍

@freesteph can you please approve again? the only change is that I deleted the "Bundled with..." from Gemfile.lock

phantas:
Merging because it's been approved by Steph, who is a member of the team and has just been added to the openregisters org

PR #664, opened 2020-02-10
Fix/update dcms register

freesteph:
### Context
We were asked to deprecate some of our registers and point to their new home.

Add a very ad-hoc solution as we were short on time and out of maintainers: add a hash mapping registers IDs with an URL to point to. If the register identifies as deprecated, some extra bit of templating are rendered.

### Guidance to review
Clone locally and add a `'prison-estate'` hash key, mapping to any URL, then visit the fake Prison Registers to see the UI in action.


app/views/registers/show.html.haml
```diff
@@ -1,4 +1,4 @@
-- @page_title = @register.safe_name
+- @page_title = @deprecated.nil? ? @register.safe_name : "[Deprecated] #{@register.safe_name}}}}}"
```
robinmitra: Too many closing braces - accidental may be?

robinmitra:
LGTM apart from my above comment.

freesteph:
Gareth & Steve are both happy with it 👍

PR #668, opened 2020-02-11
fix: properly ignore the bundle folder created by CI

freesteph:
We're suffering Bundler loosing path to gems on CI which would
indicate that the deploy script (powered by CF) is getting confused by
the artifacts of the CI build (i.e a `.bundle` folder).

This directory should be ignored by CF, however a leading slash is not
guaranteed to work in all configurations so favour `.bundle` instead.

### DIFF
--- a/.cfignore
+++ b/.cfignore
@@ -5,7 +5,7 @@
 #   git config --global core.excludesfile '~/.gitignore_global'
 
 # Ignore bundler config.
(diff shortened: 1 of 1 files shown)


freesteph:
`/.FILE` is relative to the `.gitignore` file (= `.FILE`) so this won't solve much.

PR #683, opened 2020-04-01
Fix/docker compose setup

freesteph:
### Context
The Docker Compose setup wasn't working. PostgreSQL isn't happy with an empty password, and the `sed` hackery going on in the Dockerfile means the DB configuration can take different values when running at the `docker build` time or when running the containers and mounting one's dev directory to the app to allow rapid changes.

It was also missing NPM which isn't provided by default in the `nodejs` package.

* fix the docker-compose build and allow host mounting;
* upgrade Ruby to latest version.

### Guidance to review
- as always: read the commit history;
- try locally via Docker, follow instructions in README.md


Dockerfile
```diff
@@ -1,10 +1,16 @@
-FROM ruby:2.5.5
+FROM ruby:2.6.5
+
+ENV DB_PASSWORD openregister
+ENV DB_USER openregister-info
+ENV DB_DB openregister-info_development
+ENV DB_HOST db
```
phantas, line 6: Maybe I am missing something, but why not move this to docker-compose.yml too?

freesteph: I think I was looking at the [admin Dockerfile](https://github.com/alphagov/govwifi-admin/blob/master/Dockerfile) which encodes them in there but I think you're right it makes more sense to have them in the Compose file. Have pushed a commit.

PR #684, opened 2020-04-02
Feat: offer better guidance in the support pages

freesteph:
### Context
We are getting a lot of queries coming through from members of the public worried about COVID-19 or other matters unrelated to Registers.

This PR solves this by explicitly asking users whether they work for the government or if they are a member of the public (the latter being selected by default). 

#### Civil service agents
If they identify as civil service agents present them the usual feedback form with a new "Subject" field which allows them to input more information, as opposed to the previous scenario where we would fill the subject for them: if you clicked "Question" the subject was "[Question or Feedback]", if you clicked "Problem" the subject was "[Problem]"). As we can see on the following screenshot, it is wildly unhelpful in practice:

registers_support

With this PR, we remove that vague distinction and ask them to enter it themselves, which will help triage issues faster.

#### Members of the public
If they identify as members of the public present them a page that clearly states Registers is solely for use by people who work for the government, and present a list of common issues (top one being COVID-19) and the relevant links for them to get help. This is modelled after [the work done in Notify](https://github.com/alphagov/notifications-admin/pull/3386). 

----
## Before
Two flows: support or question.

registers-current-support

## After
Two flows: member of the public or civil service agent.

registers-help-page

Notice that the "member of the public" is selected by default.

### Guidance to review
- Please try the PR locally and make sure these two flows work as expected.


config/locales/en.yml
```diff
@@ -20,6 +20,27 @@
...
+          - What you need to do
+          - https://www.gov.uk/coronavirus
+        -
+          - Contact the government
+          - Ask about benefits, driving, transport, tax, and more
+          - https://www.gov.uk/contact
+        -
+          - Avoid and report internet scams and phishing
+          - Advice on suspicious emails and text messages
+          - https://www.gov.uk/report-suspicious-emails-websites-phishing
```
quis, line 43: We included this link because a lot of the questions we got were from people Googling ‘gov sms’ or similar. I suspect fewer people with those kinds of questions are coming to Registers.

So I’d suggest:
- replace it with something more specific to the questions the public ask via Registers support
- if there’s nothing more specific to Registers then leave it, because 3 links feels better than 2

config/locales/en.yml
```diff
@@ -20,6 +20,27 @@
 # available at http://guides.rubyonrails.org/i18n.html.
 
 en:
+  support:
+    group:
+      gov: I work in the public sector and need help with the Registers API
+      public: I'm a member of the public with a question for the government
```
quis, line 26:
```suggestion
      public: I’m a member of the public with a question for the government
```

PR #696, opened 2020-06-16
Fix/gems update

freesteph:
### Context
A bunch of dependencies are outdated. They're now all up to date except `govuk-lint` which needs its own PR.

- new Makefile to automate some of these commands;
- updated `Gemfile.lock`;
- update stubbing syntax.

### Guidance to review
- review commit history;
- run app locally and make sure it works for you too.


alan-gds:
I'm afraid I can't review this, just like I can't review any Registers PRs - nothing works for me locally.  In the case of running directly, it's the bundler/rbenv/OSX issue.  In the case of docker, I'm not sure yet - that reminds me, I need to look at the logs like you suggested.

PR #713, opened 2020-08-19
Fix/error pages

freesteph:
### Context
We are getting intermittent failures from the `registers` which can be crashed by visiting any error-causing page (like a 404), because our error pages are missing a method needed to render their layout. This might possibly push the app outside of its resources, which causes a chain of crash/restarts to be triggered. The full explanation is in commit https://github.com/openregister/registers-frontend/commit/bcba8859afa640de64bbe9d7a85881b1a5c4e40a.

* Remove the hack to remove the cookie banner;
* Disable the cookie banner with a simple window-level attribute.

### Guidance to review
* Read commit log;
* Run `make serve` locally  (looks fine on my machine).

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -11,12 +11,5 @@ private
   def current_user
     @current_user ||= User.where(id: session[:user_id]).first if session[:user_id]
   end
-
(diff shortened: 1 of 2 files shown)


alan-gds (approved):
LGTM

PR #714, opened 2020-08-19
Fix/cookie banner

freesteph:
### Context
The previous PR (https://github.com/openregister/registers-frontend/pull/713) attempted to remove the cookie banner by setting a property on the global `window` object. Unfortunately this object is just set regardless by the `govuk-template` JS, then immediately called, effectively ignoring whatever value we've set[1].

So instead, just set the `seen_cookie_banner` before everyone else gets a change to, effectively discouraging `govuk-template` from showing the banner[2].

Only cavehat: the error pages still have it. See commit log for details.

\[1\]: https://github.com/alphagov/govuk_template/blob/master/source/assets/javascripts/core.js#L6
\[2\]: https://github.com/alphagov/govuk_template/blob/master/source/assets/javascripts/cookie-bar.js#L12

* Remove tentative JS fix for cookie banner;
* Enforce `seen_cookie_banner` cookie.

### Guidance to review
* Read commit log;
* Try locally (works here).

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+node_modules
+log/*
--- a/Makefile
+++ b/Makefile
@@ -8,5 +8,15 @@ populate:
 serve: build
 	docker-compose up -d
 
+stop:
+	docker-compose down
+
+# starts the web process interactively to attach to debugger breakpoints
+istart: build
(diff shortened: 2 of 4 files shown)


alan-gds (approved):
LGTM

PR #720, opened 2020-09-23
fix: add an accessibility statement

freesteph:
### Context
Add an accessibility statement.

Add a new page with the supplied content.

### Guidance to review
- clone locally;
- sense check.


alan-gds (approved):
LGTM

